PR #11, opened 2016-09-24
Fix opam package 'process' version

beurdouche:
I couldn't push that directly...


msprotz:
I thought you were in team FStarTeam that has access to kremlin?!

PR #18, opened 2017-01-17
Implement 'print_bytes' in kremlib

beurdouche:


beurdouche:
@protz Can you please check that it's ok with you and merge ? :) Thanks

jkzinzindohoue:
Would you mind moving it to testlib rather than kremlib ?
I think that as a general policy such functions that are mostly for debugging should not go to the main library. Here I am not sure the stack effect with the heap equality really account for the IO side effect.
Which I think is perfectly fine for debugging, but I would rather keep kremlib with only the essentials, and put new things in testlib or elsewhere. @protz thought ?

msprotz:
Well in that case `print_string` should also go in the testlib, right?

`print_string` just exposes `printf`; `print_bytes` is a more sophisticated function; perhaps `C.fst` should only about those things that already exist in C's stdlib?

beurdouche:
> Well in that case print_string should also go in the testlib, right?
> 
> print_string just exposes printf; print_bytes is a more sophisticated function; perhaps C.fst should only about those things that already exist in C's stdlib?
> 
Indeed, I also think that it would be better to have stdlib only in C.fst.

jkzinzindohoue:
This is something I discuss quickly with @beurdouche and @prosecco but I wish the side effect modelization was cleaner. If C.fst is meant to expose the standard library, they I believe it should follow the same naming convention etc.
And we can't have functions which post-condition is "h0==h1" if we want to account for the side effect, unless it is for debugging.

protz:
At least the function is in the Stack effect -- if it was Tot it would 
be eliminated by F*.

kremlib/C.fst
```diff
@@ -58,6 +58,11 @@ assume val print_string: string -> Stack unit
   (requires (fun h -> True))
   (ensures (fun h0 _ h1 -> h0 == h1))
 
+assume val print_bytes: b:buffer UInt8.t -> len:UInt32.t{length b >= UInt32.v len} -> Stack unit
```
beurdouche: Ok, done.

beurdouche:
Ok, I pushed the stylistic change but I let the function here so that we can move everything not part of stdlib at the same time.

msprotz:
Thanks! When you have time, happy to review a pull request for moving non-C parts into another lib.

Also, have you provided an OCaml implementation of this function? `C.ml` now lives in `kremlib/`

beurdouche:
No I haven't implemented in ML. Should I just define it as `print_bytes` the same way you did for `print_string` ? (It seems to be in the Pervasives module).
I'll submit a PR for the move as soon as we have a clean proposal ;)

msprotz:
It doesn't print hex-encoding, though, so you may want to re-implement it slightly differently

beurdouche:
> It doesn't print hex-encoding, though, so you may want to re-implement it slightly differently
> 
Ok, I will.

PR #41, opened 2017-03-13
Load/Store for 16bits integers

msprotz:


msprotz:
I merged this, but I am growing increasingly concerned that we have a bunch of macros for endian-ness that we are just NEVER testing on big endian platforms.

Perhaps it's time to have a PowerPC machine somewhere running CI, too.

Issue #45, opened 2017-06-22
Pass -fstack-check to gcc and clang


andreas23:
On Linux (and other UNIXen), if the stack grows until it overlaps with the heap, aliasing between stack and heap occurs, which leads to memory corruption and exploitable security vulnerabilities. The default mechanism on Linux to prevent that is based on a gaurd page of 4k size. An attacker can make a stack allocation jump over the guard page, as demonstrated recently: https://www.qualys.com/2017/06/19/stack-clash/stack-clash.txt

gcc provides a mostly working (but bugs are being worked on) protection against it when passed the -fstack-check argument. It is unfortunately not the default. clang currently implements the flag as a no-op, but work is on the way to do it properly.

kremlin should always pass -fstack-check to both gcc and clang.

PR #46, opened 2017-06-22
Pass -fstack-check to gcc and clang. This protects against aliasing o…

andreas23:
…f stack and heap. Implements issue #45


msprotz:
Thanks! This looks good.

PR #47, opened 2017-06-22
Fix UInt128 eq_mask to always return 2^128-1 or 0

tchajed:
The masks from both halves of the int should be combined so that eq_mask
always returns all 1s if the numbers are equal or all 0s otherwise.


jkzinzindohoue:
Thanks. You would not be willing to implement those in F* by any chance ?

tchajed:
I'm currently re-implementing the UInt128 functions in `kremlib.h` in F* and verifying them. When that's done we should be able to rely on verified F* implementations rather than anything in `kremlib.h` for UInt128.

msprotz:
Thanks Tej!

PR #61, opened 2017-10-03
Add Solaris defines for Endian-ness macros

franziskuskiefer:
kremlib.h is missing defines for endian-ness on Solaris.


msprotz:
I have no Solaris machine to try it out, so I'm going to trust you on this one... :-) thanks!

PR #62, opened 2017-10-06
Add FreeBSD support

franziskuskiefer:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1396301
The patch in https://bugzilla.mozilla.org/show_bug.cgi?id=1396301 doesn't look correct (it would brake Windows). But this should do the trick.

### DIFF
--- a/kremlib/kremlib.h
+++ b/kremlib/kremlib.h
@@ -228,6 +228,12 @@ typedef const char *Prims_string;
 #define htobe32(x) BE_32(x)
 #define be32toh(x) BE_IN32(x)
 
+/* ... for the BSDs */
+#elif defined(__FreeBSD__) || defined(__NetBSD__) || defined(__DragonFly__)
+#include <sys/endian.h>
+#elif defined(__OpenBSD__)
+#include <endian.h>
+
 /* ... for Windows */
 #elif (defined(_WIN16) || defined(_WIN32) || defined(_WIN64)) &&               \
     !defined(__WINDOWS__)


kremlib/kremlib.h
```diff
@@ -228,6 +228,12 @@ typedef const char *Prims_string;
 #define htobe32(x) BE_32(x)
 #define be32toh(x) BE_IN32(x)
 
+/* ... for the BSDs */
+#elif defined(__FreeBSD__) || defined(__NetBSD__) || defined(__DragonFly__)
+#include <sys/endian.h>
+#elif defined(__OpenBSD__)
+#include <endian.h>
```
jbeich: @landryb, do you have `<endian.h>` on OpenBSD? [`/usr/src/include`](https://github.com/openbsd/src/tree/master/include/) suggests otherwise.

msprotz:
Thanks @franziskuskiefer. There's another pull request in hacl-star about fixing kremlib.h for platforms that don't have any sort of built-in byte swapping operations, so I'm contemplating adding an #else after the last #elif to have a fallback in any case. What do you think?

kremlib/kremlib.h
landryb: `usr/include/endian.h` is a symlink to `usr/include/sys/endian.h` (cf https://github.com/openbsd/src/blob/master/sys/sys/endian.h) which says the following:
```
 * Public definitions for little- and big-endian systems.
 * This file should be included as <endian.h> in userspace and as
 * <sys/endian.h> in the kernel.
```

franziskuskiefer:
I think that would be great. Having a system independent fallback would make life much easier. That way tier-3 platforms wouldn't break.

beurdouche:
@msprotz I have hold on the HACL* PR in question for that exact reason, I would be in favor of doing what you suggest... Let me know how you want to proceed

msprotz:
Thanks... let me add fallback to kremlib.h now -- @franziskuskiefer are there any tier3 platforms we should look into?

franziskuskiefer:
> Thanks... let me add fallback code to kremlib.h now -- @franziskuskiefer are there any tier3 platforms we should look into?

I think BSD and Solaris are the most used tier3 platforms for NSS. But I'd suggest using the generic approach from 276bbd4ee67414109b5e88627cd7e2eda4eab73e for other platforms.

PR #65, opened 2017-10-25
Simplify the chain of ifdefs

msprotz:
@franziskuskiefer can you give this a try?
- The Windows #ifdefs were sketchy: BYTE_ORDER, as far as I know, was defined only in `netsysio.h`, which we were not including, and even then, was defined to be either `_BIG_ENDIAN` or `_LITTLE_ENDIAN` (note the leading underscore), meaning that the #ifdef tests were never executing right ... but apparently, the preprocessor's behavior for #if FOO == BAR when neither BAR or FOO is defined, is to say true, so we were effectively picking LITTLE_ENDIAN... I removed these tests since we don't intend to target Xbox 360
- I also flattened the nested windows tests to either test for msvc (which implies windows), or _WIN32 and clang/gcc ... I wouldn't know what to do in other cases
- the last #else of the (now flat) sequence of cascading #ifdefs errors out

Happy to iterate... I tested as best as I can, but happy to clean things up if needed.

### DIFF
--- a/kremlib/kremlib.h
+++ b/kremlib/kremlib.h
@@ -18,19 +18,19 @@
 #include "gcc_compat.h"
 
 #ifdef __GNUC__
-#define inline __inline__
+#  define inline __inline__
 #endif
 
 /* GCC-specific attribute syntax; everyone else gets the standard C inline
  * attribute. */
 #ifdef __GNU_C__
-#ifndef __clang__
-#define force_inline inline __attribute__((always_inline))
+#  ifndef __clang__
+#    define force_inline inline __attribute__((always_inline))
+#  else
+#    define force_inline inline
+#  endif
 #else
-#define force_inline inline
-#endif
-#else
-#define force_inline inline
+#  define force_inline inline
 #endif
 
 /******************************************************************************/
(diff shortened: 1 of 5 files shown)


msprotz:
Build failure is due to the upgrade of the wasm package that contains a non-backwards compatible change.

msprotz:
CC @beurdouche

kremlib/kremlib.h
```diff
@@ -370,9 +345,11 @@ typedef const char *Prims_string;
   }))
 #define be64toh(x) (htobe64((x)))
 
-#endif
+/* ... fail */
+#else
```
franziskuskiefer: Unrelated to this PR. I generally find it very helpful to put some comment to `#else`, `#endif`, and `#elif` explaining what it belongs to. When there's no indentation these constructs are always a little hard to parse imho.

msprotz:
Thanks... I've changed my clang-format option to indent #defines. I generally try to comment the #endif (see bottom of FStar.h) but for series of cascading #elifs I'm not sure what sort of comment I should put.

If you have success with this change, I'll merge... thanks!

msprotz:
Alright, thanks for the review!

PR #66, opened 2017-11-01
Add support for mutually recursive types and functions 

jroesch:
@nikswamy @protz carved out a subset of the changes, still working on the PR a bit, will update this evening.

### DIFF
--- a/Tests.ml
+++ b/Tests.ml
@@ -4,7 +4,7 @@ let _ =
   let open Constant in
   let open PPrint in
   let p_t t =
-    let d: CStar.decl = Type ("t", t) in
+    let d: CStar.decl = Type ("t", t, false) in
     let d = CStarToC.mk_stub_or_function d in
     match d with
     | Some (C.Decl ([], d)) ->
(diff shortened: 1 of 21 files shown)


msprotz:
Thanks! I've implemented the suggested changes in 7c0879be7ddd683e43027734368c60d93ffe3823

The approach I've chosen is slightly different, and is explained at https://github.com/FStarLang/kremlin/blob/fstar-master/src/DataTypes.ml#L36 -- essentially, I see a program's types as a graph, where nodes are pairs of (type, effective arguments) (e.g. `"list", [ Int ]`) and edges are just "uses" of a given type instance from another type's definition (e.g. edge from `t, [ Int ]` to `u, [ Int32 ]` if the definition of `t, [ Int ]` mentions `u Int32`.

Effectively compiling these then simply amounts to a traversal of the type graph, where cycles are broken by inserting a forward declaration. See https://github.com/FStarLang/kremlin/blob/fstar-master/test/DataTypesMut.fst#L11 for a pretty example.

PR #67, opened 2017-11-03
Fix inclusion of headers (text) in .c files

beurdouche:
The command line says the header should be included both in `.c` and `.h` files.
This is what we want but was currently done only for `.h` files.

This patch is an attempt to fix this issue :)


msprotz:
Thanks! Pushed essentially the same fix on master

PR #70, opened 2017-11-23
Fix Solaris defines for Endian-ness

psumbera:
Don't use variants with IN which expect pointer.

Sorry for the issue. Please see:

https://bugzilla.mozilla.org/show_bug.cgi?id=1405268
https://bugzilla.mozilla.org/show_bug.cgi?id=1419009

Now it should be hopefully right. I have created test program for these macros and I'm getting the same results as on Linux.


beurdouche:
Thanks ! :) I'll propagate that to HACL* and NSS asap and comment in Bugzilla when done.

PR #75, opened 2018-01-30
Total loops

tahina-pro:
A total "do-while" loop to be extracted to C. This could help avoid the use of total recursive functions in Low*.


msprotz:
Can you re-run the CI? It's back and working now.

PR #76, opened 2018-01-30
Kremlin cross-compiles

wintersteiger:
Not intended to be merged as is, just a demo to show that it doesn't take all that much. I just ran `CROSS=yes make` and, on a proper Ubuntu machine (no WSL) it gets me very swiftly to 
```
...:~/kremlin$ file _build/src/Kremlin.native 
_build/src/Kremlin.native: PE32+ executable (console) x86-64, for MS Windows
```

Those ppx options are a bit of a mess, I'm sure there's a nicer way to get those options through (e.g. in `_tags`, but I don't know how to get `$(OCAMLLIB)` into `_tags`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,17 @@
 .PHONY: all clean test
 
+ifdef CROSS
+  OCAMLLIB=$(abspath $(shell ocamlfind printconf path))
+  OCAMLBUILD=ocamlbuild -toolchain windows -I src -I lib -I parser -use-menhir -use-ocamlfind -classic-display \
+   -menhir "menhir --infer --explain" \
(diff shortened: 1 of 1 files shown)


msprotz:
I thought `ppx_deriving` was supposed to be able to deal with these cases. Perhaps worth pinging the ppx_deriving folks to see if they're missing support for cross-compilation? In any case, good news!

wintersteiger:
That would be nice if they could support cross-compilation out of the box! I must admit, I don't fully understand how it's supposed to work and I may have set up the ppx_deriving package(s) the wrong way. If those object files are not needed at link time, the *-windows package could just be a link to the build system's (non-windows) packages. I think that's pretty much the effect of those options I set.

PR #80, opened 2018-02-10
in-place byte slices

fournet:
Pls check this is safe.


BarryBo:
Safe for miTLS's usage, where we do not free the input bytes or mutate them (the caller has just allocated the input data and only passes it to slice()).

For non-miTLS, where the input bytes could be freed or mutated, this is not safe.

fournet:
I believe this is safe for `FStar.Bytes`, whose C implementation already relies on bytes not being freed or mutated, not just for its use in miTLS.

msprotz:
Can you remove the commented out code from the source tree? We can always recover it via version control.

Thanks for the pull request!

@BarryBo do we know that this significantly reduces the amount of allocations?

PR #81, opened 2018-02-18
Android/x86 requires limits.h to use SIZE_MAX macro

makotokato:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1439226

Since this library does't include limits.h, SIZE_MAX is undefined.  So it requires this header.


makotokato:
I should file PR for fstar-master.  So I will create new PR.

PR #82, opened 2018-02-19
Android/x86 requires limits.h to use SIZE_MAX macro

makotokato:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1439226

Since this library does't include limits.h, SIZE_MAX is undefined. So it requires this header.

### DIFF
--- a/kremlib/kremlib.h
+++ b/kremlib/kremlib.h
@@ -2,6 +2,7 @@
 #define __KREMLIB_H
 
 #include <inttypes.h>
+#include <limits.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>


msprotz:
Thanks! I take it that limits.h exists on all platforms, Windows included?

I can cherry-pick on `master` if Mozilla is stuck with an old version of kremlib.h

Thanks,

Jonathan

makotokato:
Thanks.  I am happy if you merge this change to master too.

Also , although I check VS2017 headers, SIZE_MAX is defined by stdint.h and limits.h (C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.12.25827\include)

franziskuskiefer:
> I can cherry-pick on master if Mozilla is stuck with an old version of kremlib.h, let me know

We should get this into HACL*. But we're currently blocked by landing the vectorised ChaCha20 code from HACL* in NSS before we can take anything else. That should happen soon though.

msprotz:
Alright, thanks. I've cherry-picked on master, in case that's useful.

PR #84, opened 2018-03-28
Avoid sscanf(), which doesn't work well for uint8_t on Windows

BarryBo:
Also, use safecrt on Windows when possible... part of mitls-papers issue # 25.

This seems simpler/faster/safer than continuing to use sscanf().  This code runs during miTLS startup, to parse a long crypto key which is initialized from a string.

The printing code is live in our production build, called from the tracing routines.


BarryBo:
@msprotz When you're back, the second commit resolves the TestKremBytes failure.  There was a sign extension bug in the middle of several other bugs, some dumb, some subtle.

PR #85, opened 2018-04-10
Loop unrolling for Kwasm

denismerigoux:
Kwasm now has the same unrolling feature than Cwasm, triggered by `funroll-loops`.

The indentation changes come from ocp-indent on `AstToCFlat.ml`. I can also leave it the way it was before.

### DIFF
--- a/src/AstToCFlat.ml
+++ b/src/AstToCFlat.ml
@@ -525,6 +525,42 @@ and mk_expr (env: env) (locals: locals) (e: expr): locals * CF.expr =
       ) locals branches in
       locals, CF.Switch (e, branches, default)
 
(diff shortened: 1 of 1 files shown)


msprotz:
Hey Denis, thanks for the pull request. Please, no re-indentation, I can't see what you changed. If we want to switch to ocp-indent later, let's do it in one big pass.

I would't reuse -funroll-loops which is already user for loops with constant bounds; can you use -finline-functions (also a GCC option, the name is standard)?

Thanks,

Jonathan

denismerigoux:
The indentation is fixed. Sorry for the confusion, I do mean *loop unrolling* and not inlining, so the name of the option is adapted (and already used for loop unrolling in the C backend).

PR #90, opened 2018-05-01
C89 Compatibility

wintersteiger:
Some minor tweaks to make kremlib more palatable to C89 compilers and VS 2010.

### DIFF
--- a/kremlib/krembytes.h
+++ b/kremlib/krembytes.h
@@ -3,8 +3,8 @@
 #ifdef __FStar_H_DEFINED
 #define __KREMBYTES_H
 
-#include <inttypes.h>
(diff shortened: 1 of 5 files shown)


msprotz:
Thanks. Antoine requested a restructuring of kremlib so I'm going to hold off on the merge of this one, also on the basis that some compilers don't necessarily support things such as "%hh", so un-conditionally defining this macro is not always the right thing to do.

wintersteiger:
Yeah, I only added those bits that were used in the rest of the code, but I think we should try to minimize what we actually use, so that we don't have to re-implement `inttypes.h` for all platforms...

PR #95, opened 2018-05-22
Removed designated initializers when -fc89 is on.

wintersteiger:
When `-fc89` is on, designated initializers are not removed from the output. I thought a couple weeks ago we got around this by using `-fnocompound-literals`, but that didn't work.


msprotz:
Hey, thanks for the pull request. Did you extensively test this to make sure that the fields are always in the right order? As far as I recall, I never checked exhaustively that KreMLin preserves the order of the fields (which is why I like designated initializers!).

msprotz:
Also: I pushed a fix for CI on master which you should be able to merge into your branch, to get a green too.

wintersteiger:
Nope, I fully relied on you having checked that already! Are there any unordered data-structures involved with tracking field names/types?

msprotz:
There aren't, I'm just wary of some `List.rev` somewhere.

I'll double-check and merge. Thanks!

msprotz:
This patch does not work in the case of a C union, where the designated initializer is essential to selecting the case of the union that we want.

This generates the following error in miTLS:

```
Cert.c:272:13: error: incompatible pointer to integer conversion initializing 'TLSError_alertDescription' with an expression of type 'Prims_list__FStar_Bytes_bytes *'
      (aka 'struct Prims_list__FStar_Bytes_bytes_s *') [-Werror,-Wint-conversion]
          { buf }
```

The solution is to desugar compound literals into an allocation followed by a series of assignments, which I thought -fcompound-literals did already. It's quite possible that -fcompound-literals does not work as intended, or is incomplete, and I think that would be the right way to fix this issue.

msprotz:
The error above stems from the following type definition:
```
 152 typedef struct
 153 FStar_Error_optResult__K___TLSError_alertDescription_Prims_string_Prims_list__FStar_Bytes_bytes_s
 154 {
 155   FStar_Error_optResult__K___TLSError_alertDescription_Prims_string_K___FStar_Bytes_bytes_FStar_Bytes_bytes_tags
 156   tag;
 157   union {
 158     K___TLSError_alertDescription_Prims_string case_Error;
 159     Prims_list__FStar_Bytes_bytes *case_Correct;
 160   }
 161   val;
 162 }
 163 FStar_Error_optResult__K___TLSError_alertDescription_Prims_string_Prims_list__FStar_Bytes_bytes;
```
the C compiler does not know, lacking a designated initializer, which one of the two cases is being initialized. (Or at least that's my understanding of it.)

msprotz:
Here's a minimal test:
```
module Test

type point3d = { x: Int32.t; y: Int32.t; z: Int32.t }
type point = { x: Int32.t; y: Int32.t }

type t =
  | A of point
  | B of FStar.Buffer.buffer point3d

let x = B (C.Nullity.null point3d)
```
compile with `-warn-error +9` and this feature activated.

wintersteiger:
Bump! I still need this for the hacl into mbedTLS integration. I use my patch for now and I only need this for `FStar_UInt128.c` and `Hacl_Curve25519.c`; if those don't use unions we should be fine. 

I agree though, we should fix this more generally, I just don't have the time right now though.

msprotz:
I am reluctant to merge a feature knowing that it only works in a few cases and is broken the rest of the time. It looks like we're both short on time, and this is non-trivial. What needs to happen is:
- detect if the type contains a union
- if it doesn't, just omit the designated initializers
- if it does, hoist the structure as an uninitialized let-binding in the enclosing scope, and emit a series of assignments
I think it's reasonable that for the time being you maintain that one-liner in your local setup, and it will be superseded by a fix for this issue when I get to it.

Thanks,

Jonathan

wintersteiger:
Alright, I think I have an acceptable fix now. Whenever `-fc89` is on, it omits the designation for structs, as well as for unions if the initialized field is the first one. For those cases where the designator is needed, it throws a warning and keeps the designator, so that the C compiler will complain (or it swallows it anyways, like gcc). I did not go into re-ordering the union members or adding a global initialization function etc.

This is all I need for now, and I think it's safe to use generally until someone needs more (and in those cases it may be easier to not produce unions that require it, internally).

msprotz:
The commit range 755768cda43ee19768f497400e2c95d422803e3e..04e501add373c3c94f0645ee8f00714e9ac63db3 fixes the compilation of struct literals when `-fno-compound-literals` is used. This gives progressive initialization using a temporary, and I was able to confirm (via the new distribution, now under CI, of hacl-star/dist/compact-c89) that the code can be successfully compiled with `-std=c89`.

The fix was much deeper and could not be performed at the level of the final pretty-printing phase, so I'm closing this. Hopefully you won't need a patch anymore to generate proper c89 code.

PR #96, opened 2018-05-22
Added casts to squelch warnings emitted by VC2010.

wintersteiger:
Should be syntatic only.

### DIFF
--- a/include/kremlin/fstar_ints.h
+++ b/include/kremlin/fstar_ints.h
@@ -19,11 +19,11 @@ typedef int8_t FStar_Int8_t, FStar_Int8_t_;
 
 static inline uint32_t rotate32_left(uint32_t x, uint32_t n) {
   /*  assert (n<32); */
-  return (x << n) | (x >> (-n & 31));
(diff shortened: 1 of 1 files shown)


msprotz:
This looks good to me but these things are tricky and I'd like to get a second pair of eyes to confirm we're not tripping on some corner case. @tahina-pro can you confirm that per the C standard, this doesn't change the semantics of the code?

I particular: for n of type uint32_t, are `-n` and `-(int32_t)n` the same thing? (i.e. in the first case is the integer promotion rule stating that `n` gets converted to signed without an extension?)

It shouldn't matter since n is less than 32 in any case, but still...

wintersteiger:
Good point! Technically, this may be undefined behavior in some versions of the C standard. We can turn `-n` into `(~n + 1)` to make it obvious what should happen.

wintersteiger:
@tahina-pro just checking, did you have a chance to look at this?

msprotz:
Closing this because I've removed the rotate functions. Seeing how easy these are to get wrong, didn't want to keep these in my tree :)

PR #101, opened 2018-06-13
Allow reading KRML_HOME from environment

pierrebeaucamp:
Hello,

this is a small patch I implemented for my local installation of kremlin to allow reading `KRML_HOME` from the environment - similar to how that is possible with `FSTAR_HOME`. I needed this because the assumption that `KRML_HOME` is always in the grandparent directory of the binary doesn't hold true in my environment.

I understand that this might be too specific to my setup as it doesn't seem to cause any problems for others. However I thought the patch might be helpful for various other reasons, so I decided to open a PR.

I apologize if I there are any coding errors - I don't really know OCaml, so I don't expect this to be bullet-proof. I mainly based it on how the `FSTAR_HOME` variable is read in the same file.

I could also not find any contribution guidelines, so I don't know if this PR or the commit messages are acceptable. Please let me know if I should change anything

### DIFF
--- a/src/Driver.ml
+++ b/src/Driver.ml
@@ -142,8 +142,13 @@ let detect_kremlin () =
     KPrint.bprintf "%sthe Kremlin executable is:%s %s\n" Ansi.underline Ansi.reset real_krml;
 
     let krml_home =
-      try read_one_line !readlink [| "-f"; d real_krml ^^ ".." ^^ ".." |]
-      with _ -> fatal_error "Could not compute krml_home"
+      begin try
+        Sys.getenv "KRML_HOME"
+      with Not_found -> try
+        read_one_line !readlink [| "-f"; d real_krml ^^ ".." ^^ ".." |]
+      with _ ->
+        fatal_error "Could not compute krml_home"
+      end
     in
     KPrint.bprintf "%sKreMLin home is:%s %s\n" Ansi.underline Ansi.reset krml_home;
 


msprotz:
Thanks for the patch, everything looks fine. Good point about CONTRIBUTING.md missing, I'll add something.

I'm curious about your PR -- are you just moving around the kremlin binary? In that case, would the following work for you:
```
PREFIX=$HOME/opt/kremlin make install
```
?

pierrebeaucamp:
Hm, I'm already setting that prefix when installing. My problem is that I'm running on NixOS and therefore I don't really have those "nice" paths like `/opt/kremlin` or `/usr/local/bin` etc. My install prefix is `/nix/store/kjws66m32i1j617vfz2pab9j3n369i0h-kremlin`, and `make install` places the binary in `/nix/store/kjws66m32i1j617vfz2pab9j3n369i0h-kremlin/bin/krml`.

Calling `../..` on that path sets the default `KRML_HOME` to `/nix/store`, which is obviously wrong. To me, the simplest solution seemed to allow setting `KRML_HOME` as an environment variable, similar to fstar (where I had exactly the same problem).

msprotz:
Thanks for the context. I was wondering if it was worth having KREMLIB and KREMINC variables that are separate for the kremlib/ and include/ directories, but this use-case should already be taken care of by the PREFIX feature of the Makefile, which I understand you won't need in your case.

Looks good to me, thanks for the patch!

PR #103, opened 2018-07-14
Don't overwrite OCAMLPATH in Makefile

pierrebeaucamp:
This is exactly the same issue as https://github.com/FStarLang/FStar/pull/1456

Essentially, instead of overwriting the entire `OCAMLPATH` to include the fstar library, we just append it instead.


msprotz:
Looks good, thanks!

msprotz:
I had to revert this because it broke our windows tests: even on cygwin, OCAMLPATH is semicolon-separated. I'll push a fix later.

pierrebeaucamp:
Ah sorry, I did not know this. I assumed OCAMLPATH followed similar behavior as the Unix PATH. Sorry for breaking things

msprotz:
No worries, I didn't notice either until CI ran

PR #114, opened 2018-10-08
Added copyright header to kremlib files.

wintersteiger:
We need to add some sort of license information to the kremlib files if we are to distribute them in part and in snapshots, etc. See example; this is the format recommended by the Microsoft OSS team.

- Not sure if we need to include INRIA here, but I guess it's a good idea anyways. 
- I also want copyright headers in all extracted files and I'm thinking about adding a flag to kremlin to automate that during extraction. Probably overkill and yet another thing to maintain, but it would make my life easier.


msprotz:
the `-header` option already allows you to customize the initial comment block of generated files; I know that HACL* uses it

msprotz:
the copyright block seems fine for me; can you extend your pull request to everything that's in kremlib/ and possibly src/? or is it too much and we just want the distributed files?

wintersteiger:
Alright, that should suffice!

wintersteiger:
Sure, no problem, that should be doable quickly with a bit of scripting.

wintersteiger:
`-header` works, but it replaces the ''This file was generated by KreMLin..." messages. I think I want both, but I'm not fully decided yet. Would it make sense to add a magic `$INVOCATION` variable or some such thing?

msprotz:
@beurdouche would you mind if `-header` kept the comment that tells how the file was generated?

msprotz:
@wintersteiger : note that you added src/AutoConfig.ml to version control, even though it was in src/.gitignore -- I removed it in 5f5fb82

wintersteiger:
Thanks! I neglected to check whether it's checked in before force-pushing.

PR #118, opened 2018-11-19
Added implemenations for some FStar.Int.Cast functions.

wintersteiger:
Is there where these functions should go?


s-zanella:
I'm curious about why you need these. KreMLin already extracts `Int.Cast.uint32_to_uint64 x` as `(uint64_t)x`

wintersteiger:
I thought it would do those automatically too, but then I got undefined references. It may be because I defined `let u32_64 = Int.Cast.uint32_to_uint64` just for the shorter name.

wintersteiger:
It does, thanks!

PR #133, opened 2019-03-26
Ctypes bindings generation

msprotz:
Submitted the pull request in the wrong direction. Let's use this pull request @victor-dumitrescu


parser/KLexer.ml
```diff
@@ -1,5 +1,5 @@
 open Ulexing
-open Parser
+open Kparser
```
msprotz: Nit: can we use `KParser` to be more in line with `KList`, etc.?

src/AstToCFlat.ml
```diff
@@ -405,7 +405,7 @@ and mk_expr (env: env) (locals: locals) (e: expr): locals * CF.expr =
 
   | EBufCreate (l, e_init, e_len) ->
       if not (e_init.node = EAny) then
-        Warnings.fatal_error "init node is not any but %a (see SimplifyWasm)\n" pexpr e_init;
+        Warn.fatal_error "init node is not any but %a (see SimplifyWasm)\n" pexpr e_init;
```
msprotz: I don't know if it's feasible but is there a way to use -pack or something for compiler-libs to avoid all those pesky renamings? For instance this: https://opam.ocaml.org/packages/ocaml-compiler-libs/ seems to expose names that don't conflict.

src/GenCtypes.ml
```diff
@@ -0,0 +1,276 @@
...
+
+open Lexing
+open Pprintast
+open Ast_helper
+open Asttypes
+open Location
+open Longident
+
+
+let migration = Versions.migrate Versions.ocaml_405 Versions.ocaml_current
```
msprotz: So we are moving from the current version to the 4.05 AST. Any reason for 4.05? Can we pick something more recent, e.g. 4.08? (I don't have a strong opinion: just wondering.)

src/GenCtypes.ml
```diff
@@ -0,0 +1,276 @@
...
+
+let mk_app e args =
+  Exp.apply e (List.map (fun a -> (Nolabel, a)) args)
+
+let mk_simple_app_decl (name: ident) (typ: ident option) (head: ident)
+  (args: expression list): structure_item =
+  let p = Pat.mk (Ppat_var (mk_sym name)) in
+  let t = Option.map (fun x -> Typ.mk (Ptyp_constr (mk_ident x, []))) typ in
+  let e = mk_app (exp_ident head) args in
+  mk_decl ?t:t p e
```
msprotz: I think you can use shorthand `?t` instead of `?t:t`

src/GenCtypes.ml
```diff
@@ -0,0 +1,276 @@
...
+
+
+(* manipulating identifiers *)
+let mk_unqual_name module_name  n =
+  if KString. starts_with n (module_name ^ "_") then
```
msprotz: nit: spurious space

src/GenCtypes.ml
```diff
@@ -0,0 +1,276 @@
...
+let mk_simple_app_decl (name: ident) (typ: ident option) (head: ident)
+  (args: expression list): structure_item =
+  let p = Pat.mk (Ppat_var (mk_sym name)) in
+  let t = Option.map (fun x -> Typ.mk (Ptyp_constr (mk_ident x, []))) typ in
+  let e = mk_app (exp_ident head) args in
+  mk_decl ?t:t p e
+
+
+(* manipulating identifiers *)
+let mk_unqual_name module_name  n =
```
msprotz: nit: double space

src/GenCtypes.ml
```diff
@@ -0,0 +1,276 @@
...
+  | Struct
+  | Union
+
+let structured_kw = function
+  | Struct -> "structure"
+  | Union -> "union"
+
+let mk_struct_manifest (k: structured) t =
+  let tag = match k with
+   | Struct -> t
```
msprotz: nit: indentation

src/GenCtypes.ml
```diff
@@ -0,0 +1,276 @@
...
+  let tm = mk_struct_manifest k unqual_name in
+  let ctypes_structure =
+    let struct_name = match k with
+      | Struct -> mk_struct_name name
+      | Union -> ""
+    in
+    let t = Typ.mk (Ptyp_constr (mk_ident "typ", [tm])) in
+    let e = mk_app (exp_ident (structured_kw k)) [mk_const struct_name] in
+    let p = Pat.mk (Ppat_var (mk_sym unqual_name)) in
+    mk_decl ?t:(Some t) p e
```
msprotz: nit: shorthand `~t`

src/GenCtypes.ml
```diff
@@ -0,0 +1,276 @@
...
+let build_binding module_name name return_type parameters : structure_item =
+  let func_name = KString.chop name (module_name ^ "_") in
+  let e = build_foreign_exp module_name name return_type parameters in
+  let p = Pat.mk (Ppat_var (mk_sym func_name)) in
+  mk_decl p e
+
+let mk_ctypes_decl module_name (d: decl): structure =
+  match d with
+  | Function (_, _, return_type, name, parameters, _) ->
+      (* Don't generated bindings for projectors and internal names *)
```
msprotz: nit: generate~~d~~

src/GenCtypes.ml
```diff
@@ -0,0 +1,276 @@
...
+      if not (KString.starts_with name (module_name ^ "___proj__")) &&
+         not (KString.starts_with name (module_name ^ "_uu___"))then
+        [build_binding module_name name return_type parameters]
+      else
+        []
+  | Type (name, typ, _) -> begin (* VD: do I need to consider flags here? *)
```
msprotz: No you can safely ignore them. The only one you might possibly be interested in is `Const` but it's a poorly designed feature and no one uses it anyhow

test/ctypes/Makefile
```diff
@@ -0,0 +1,76 @@
...
+
+.PHONY: .FORCE
+.FORCE:
+endif
+
+include .depend
+
+ALL_CHECKED_FILES = $(addsuffix .checked,$(ROOTS))
+
+all: Client.exe
```
msprotz: `all` is not the default target as it does not come first... need to invoke `make all` manually

test/ctypes/Makefile
```diff
@@ -0,0 +1,76 @@
...
+	mkdir -p bindings
+	$(KRML) $(KRML_FLAGS) \
+	  -tmpdir bindings -skip-compilation $(filter %.krml,$^) \
+	  -ccopt -Wno-unused \
+	  -warn-error @4 \
+      -ctypes '$*'
+
+.PHONY: Client.exe
+Client.exe: Lowlevel.bind
+	cc -I /home/victor/github/project-everest/everest/kremlin/include -c -o Lowlevel.o bindings/Lowlevel.c
```
msprotz: use `../../include` instead

test/ctypes/Makefile
```diff
@@ -0,0 +1,76 @@
...
+	ocamlfind ocamlopt -I bindings -linkpkg -package ctypes,ctypes.foreign,ctypes.stubs -cclib -Wl,-E -cclib Lowlevel.o bindings/Lowlevel_bindings.ml bindings/Gen_stubs.ml -o bindings/Gen_stubs.exe
```
msprotz: `-E` seems to be linux-only

test/ctypes/Makefile
```diff
@@ -0,0 +1,76 @@
...
+	./bindings/Gen_stubs.exe
+	cc -I /home/victor/github/project-everest/everest/kremlin/include -I bindings -I $(CTYPES) -c -o Lowlevel_c_stubs.o bindings/Lowlevel_c_stubs.c
```
msprotz: use `ocamlopt` instead of `cc`(or actually, `$(CC)`) since it will invoke i) the right `cc` that is suitable for your toolchain and will also pass the include paths to the ocaml compiler internals (this fixed it on my machine)

test/ctypes/Makefile
```diff
@@ -0,0 +1,76 @@
...
+	ocamlfind ocamlopt -I bindings -linkpkg -package ctypes,ctypes.foreign -cclib -Wl,-E -cclib Lowlevel.o Lowlevel_c_stubs.o bindings/Lowlevel_bindings.ml bindings/Lowlevel_stubs.ml Client.ml -o Client.exe #
```
msprotz: this failed for me:
```
File "Client.ml", line 19, characters 38-49:
Error: Unbound type constructor Lowlevel.tr
```

victor-dumitrescu:
I think I tried using this package when I started this but I can't recall why I didn't in the end. I'll give it another go.

From my understanding of how the Ctypes stub generation API works, the bindings need to be declared as part of a functor, which is then passed to the Ctypes functions that generate the actual ML and C bindings at compile time. Is there another way to generate this functor that doesn't involve building abstract syntax, writing it to a file and compiling it? (Would it be feasible to metaprogram this, with something like MetaOCaml?)

src/GenCtypes.ml
victor-dumitrescu: I think it's the other way around. We're building a 4.05 AST which we then migrate to the version we're currently using. I think this could in principle work with any version, when I chose 4.05 (both here and in F*) it was because I thought it's the lowest supported version (looking at the README now I see that 4.04 is actually also supported?)

src/GenCtypes.ml
```diff
@@ -0,0 +1,280 @@
...
+  | Union
+
+let structured_kw = function
+  | Struct -> "structure"
+  | Union -> "union"
+
+let mk_struct_manifest (k: structured) t =
+  let tag = match k with
+    | Struct -> t
+    | Union -> "anonymous" (* unions only ever appear as anonymous fields in structs *)
```
msprotz: might be controlled by -fno-anonymous-unions

PR #134, opened 2019-03-26
Use -ccopts in final Makefile target

uranther:
This change allows the developer to add `-ccopts -lkremlib,-L$(KREMLIN_HOME)/kremlib/dist/generic` to their Makefile so that kremlib is linked in the final executable.

The behavior before was `cc C.o HelloWorld.o C_Endianness.o -o HelloWorld` which failed because it was missing the function body of `C_String_print` (call to undefined function), for example.

### DIFF
--- a/misc/Makefile.basic
+++ b/misc/Makefile.basic
@@ -29,6 +29,7 @@ OBJS 	+= $(patsubst %.c,%.o,$(SOURCES))
 all: $(USER_TARGET)
 
 $(USER_TARGET): $(OBJS)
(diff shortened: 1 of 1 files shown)


msprotz:
I'm closing on the basis that the intended usage is `-o foobar.exe` which will in turn activate linking against libkremlib.a -- since the tutorial has been updated recently, I'd gladly take a pull request for the tutorial to be updated and make a note of this particular caveat

Thanks,

Jonathan

Issue #131, opened 2019-03-18
Tests fail due to options conflict using hardened GCC 8.3.0


ghost:
`make test` output from the latest master:

```
⚙ KreMLin will drive the C compiler. Here's what we found:                                                                                                 
gcc is: gcc                                                                                                                                                
gcc options are: -I /home/james/Code/kremlin/kremlib -I /home/james/Code/kremlin/include -I .output/TestKremBytes.out -Wall -Werror -Wno-unused-variable -Wn
o-unknown-warning-option -Wno-unused-but-set-variable -g -fwrapv -fstack-check -D_BSD_SOURCE -D_DEFAULT_SOURCE -Wno-parentheses -std=c11                   
⚡ Generating object files                                                                                                                                  
✘ [CC,.output/TestKremBytes.out/TestKremBytes.c] (use -verbose to see the output)
cc1: error: ‘-fstack-check=’ and ‘-fstack-clash_protection’ are mutually exclusive.  Disabling ‘-fstack-check=’ [-Werror]                                  
cc1: error: unrecognized command line option ‘-Wno-unknown-warning-option’ [-Werror]                                                                       
cc1: all warnings being treated as errors 
```

After removing `-fstack-check` from `src/Options.ml` and building krml again, the tests pass.

I assume that `-fstack-clash-protection` is enabled because I am using a hardened toolchain with stack smashing protection (SSP) enabled.

```
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-pc-linux-gnu/8.3.0/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /var/tmp/portage/sys-devel/gcc-8.3.0/work/gcc-8.3.0/configure --host=x86_64-pc-linux-gnu --build=x86_64-pc-linux-gnu --prefix=/usr --bindir=/usr/x86_64-pc-linux-gnu/gcc-bin/8.3.0 --includedir=/usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/include --datadir=/usr/share/gcc-data/x86_64-pc-linux-gnu/8.3.0
[...]

PR #135, opened 2019-03-26
Remove -fstack-check in default GCC opts

uranther:
Addresses #131. 

The `-fstack-check` flag conflicts with `-fstack-clash-protection` which is included by default when GCC is built with stack smashing protection (SSP), causing builds to fail.


msprotz:
Closing this since I removed it a while ago. Thanks for the original report and suggestion that motivated me to fix it!

PR #137, opened 2019-05-30
Implement arithmetic shift right operators

s-zanella:


msprotz:
Looks good to me. Shouldn't break anything since no one uses these yet!

PR #141, opened 2019-06-26
Make some POSIX calls failsafe

tahina-pro:
I am trying to build a standalone binary package for EverParse including F* and KReMLin, that users should be able to run without OCaml, Cygwin, MinGW or anything.
With the fixes brought by this pull request, and with KREMLIN_HOME and KRML_HOME suitably set, I manage to successfully extract C code using `-skip-compilation`.


src/Driver.ml
```diff
@@ -99,17 +99,21 @@ let mk_tmpdir_if () =
 
 (** Check whether a command exited successfully *)
 let success cmd args =
-  match Process.run cmd args with
-  | { Process.Output.exit_status = Process.Exit.Exit 0; _ } -> true
-  | _ -> false
+  try
```
msprotz: can you document here what kind of errors might be thrown?

src/Driver.ml
```diff
@@ -124,7 +128,7 @@ let detect_base_tools () =
     if read_one_line "uname" [||] = "Darwin" && !readlink = "readlink" then
       KPrint.bprintf "Warning: OSX detected and no greadlink found. Suggestion: \
         [brew install coreutils]\n";
-  with Process.Exit.Error _ ->
+  with _ ->
```
msprotz: same here, since you now have a catch-all in success, what kind of exception could possibly happen here?

src/Driver.ml
msprotz: that is, beyond Process_erro

tahina-pro: `UNIX.Unix_error(20, "create_process", "which")` this is what I have if I run KReMLin outside of a Cygwin environment.

msprotz: can we catch Process_error and Unix_error (ENOENT then?
```
# ((Obj.magic 20): Unix.error);;
- : Unix.error = Unix.ENOENT
```

msprotz:
Thanks, this looks good, please merge as soon as you get a green!

PR #142, opened 2019-07-03
KRML_HOME -> KREMLIN_HOME

tahina-pro:
Nobody seems to be using KRML_HOME, but everybody in Project Everest is using KREMLIN_HOME instead.


msprotz:
Looks good to me. I assume that you've double-checked this creates no Everest regressions? I'll let you watch for any breakages.

Thanks,

Jonathan

PR #149, opened 2019-09-26
Extract top-level gcmalloc and ugcmalloc without initializers

s-zanella:


msprotz:
This looks great. And, yes, static storage is guaranteed to be zero-initialized. As long as the frontend only accepts EAny and Constant 0 for default values, then we're good.

Thanks for a great patch.

PR #154, opened 2019-11-25
Update types.h

franziskuskiefer:
The `elif` branch here also needs make sure we're on a x64 machine.


msprotz:
Thanks Franziskus, this seems sensible. Re-running CI to see if we get a green (I don't see why not).

msprotz:
Famous last words, the linux build failed. Looking into it...

msprotz:
```
[STDOUT]/home/everest/kremlin/kremlib/dist/generic/libkremlib.a(fstar_bytes.o): In function `FStar_Bytes_uint128_of_bytes':
[STDOUT]/home/everest/kremlin/kremlib/dist/generic/fstar_bytes.c:193: undefined reference to `FStar_UInt128_uint64_to_uint128'
[STDOUT]/home/everest/kremlin/kremlib/dist/generic/fstar_bytes.c:195: undefined reference to `FStar_UInt128_shift_left'
[STDOUT]/home/everest/kremlin/kremlib/dist/generic/fstar_bytes.c:196: undefined reference to `FStar_UInt128_uint64_to_uint128'
[STDOUT]/home/everest/kremlin/kremlib/dist/generic/fstar_bytes.c:196: undefined reference to `FStar_UInt128_logxor'
```

msprotz:
According to https://sourceforge.net/p/predef/wiki/Architectures/ _M_X64 is only defined by Visual Studio, meaning this check will always be false since it follows `!defined(_MSC_VER)`.

Did you intend for something like `__x86_64__`?

franziskuskiefer:
> Did you intend for something like `__x86_64__`?

yeah.... we should do `__x86_64__` or `__x86_64`. Unfortunately `build_local.sh` doesn't work for me so lets see what the CI says.

msprotz:
CI failed but that's because your branch is now slightly behind master...

PR #156, opened 2019-11-26
x64 patch for types.h and fstar_uint128.c

franziskuskiefer:
Closing #154 in favour of this one.
Not sure what's up with the Windows box.


msprotz:
Are there any platforms that define one but not the other? In other places I've done feature detection only on `__x86_64__` but I'm sure you get to check more platforms with Mozilla CI and if both macros need to be checked, let me know, as I'll update code in other places

franziskuskiefer:
Yeah we use both in NSS because some compiler define one and some the other. We actually need another one as well `defined(__aarch64__)`.

PR #157, opened 2019-11-26
Include stdint for __int128

franziskuskiefer:
stdint.h is required on older compilers to get __int128


msprotz:
Out of curiosity, which old x64 compilers need this?

franziskuskiefer:
I didn't do an exact check but everything GCC <= 4.6 [it seems](https://stackoverflow.com/questions/16088282/is-there-a-128-bit-integer-in-gcc).

PR #158, opened 2019-11-26
check for aarch64; don't use deprecate on GCC <= 4.4

franziskuskiefer:
GCC <= 4.4 doesn't support the deprecated attribute.
On aarch64 we have unint128 as well.


msprotz:
Thanks! This is very helpful.

PR #159, opened 2019-12-05
Add support for passthrough comments

tahina-pro:
This pull request supports:
1. a new surface AST construct to extract calls to `C.comment_gen` to `EComment`
2. new surface and internal AST constructs, `EStandaloneComment`, to extract calls to `C.comment_gen` to `CStar.Comment`, to support standalone comments that are not necessarily wrapped around an expression. Indeed, the sole use of `EComment` in these cases, around an unused pure expression, leads to its erasure, which `EStandaloneComment` claims to prevent.

This pull request has a Low* extraction counterpart: FStarLang/FStar#1895

### DIFF
--- a/src/PrintAst.ml
+++ b/src/PrintAst.ml
@@ -181,6 +181,8 @@ and print_expr { node; typ } =
   match node with
   | EComment (s, e, s') ->
       surround 2 1 (string s) (print_expr e) (string s')
+  | EStandaloneComment s ->
+      surround 2 1 (string "/*") (string s) (string "*/")
   | EAny ->
       string "$any"
   | EAbort s ->
--- a/.docker/build/config.json
+++ b/.docker/build/config.json
@@ -4,7 +4,7 @@
     "BaseContainerIsEverestImage" : true,
     "BaseContainerImageName" : "fstar",
     "BaseContainerImageTagOrCommitId": "latest",
-    "BranchName" : "taramana_kremlin_comments",
+    "BranchName" : "master",
     "GithubCommitUrl" : "https://github.com/FStarLang/fstar/commit",
     "OnDemandBuildDefinition" : "FStar\\FStar-{agentOS}",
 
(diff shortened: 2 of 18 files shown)


msprotz:
thanks for the pull request! quick question before I look at this in detail: why not declare that EComment nodes are *not* readonly expressions and that therefore they should not be eliminated?

tahina-pro:
I thought about it. There are definitely comments that we would like to never be erased (hence `EStandaloneComment`); besides, do we want to always preserve an expression enclosed by comments even if we know that extraction may optimize the expression away?

Your suggestion may call for an orthogonal construct ("`ENoErase`?"), to prevent an expression from being optimized away, or at least an extra argument to `EComment`; bearing in mind that those would not be enough to prevent erasure of pure unit expressions (these are already being erased by F*, not KReMLin, thinking of lemma calls and the like.)

src/PrintAst.ml
```diff
@@ -181,6 +181,8 @@ and print_expr { node; typ } =
   match node with
   | EComment (s, e, s') ->
       surround 2 1 (string s) (print_expr e) (string s')
+  | EStandaloneComment s ->
+      surround 2 1 (string "/*") (string s) (string "*/")
```
msprotz: @tahina-pro you need to use `words` here

msprotz: see http://gallium.inria.fr/~fpottier/pprint/doc/PPrintCombinators.html

msprotz:
Copy/pasting here what I wrote on slack:
> the pull request looks good; I have a slight preference for creating LowStar.Comment.fst (as opposed to adding them in the C module) -- C is kind of a catch-all and half the things in there are outdated; I'm trying to put new things in the LowStar namespace (edited) 
> but, not a dealbreaker

also: can you use `p_comment` for printing the comment?

src/PrintAst.ml
tahina-pro: this printer here is for debugging purposes only

tahina-pro:
`comment` and `comment_gen` are now in F* ulib, in the `LowStar.Comment` module, now merged to F* `master`, cf. https://github.com/FStarLang/FStar/blob/master/ulib/LowStar.Comment.fsti and the corresponding .fst

PR #160, opened 2019-12-10
Use native uint128 on Windows with clang-cl

franziskuskiefer:


msprotz:
thanks for the patch!

I think you might need to change
```
#if !defined(_MSC_VER)
```

into

```
#if !defined(_MSC_VER) || defined(__clang__)
```

a few lines below...?

franziskuskiefer:
So this should be correct now but my Visual Studio 2017 clang-cl crashes when compiling with this. I'll try 2019 later. But this should probably get some testing CI as well 😬

include/kremlin/internal/types.h
```diff
@@ -72,13 +75,13 @@ typedef FStar_UInt128_uint128 FStar_UInt128_t, uint128_t;
...
-#if !defined(_MSC_VER)
+#if !defined(_MSC_VER) || defined(__clang__)
 #include "fstar_uint128_gcc64.h"
 #endif
 
-#if !defined(KRML_VERIFIED_UINT128) && defined(_MSC_VER)
+#if !defined(KRML_VERIFIED_UINT128) && defined(_MSC_VER) && !defined(__clang__)
 #include "fstar_uint128_msvc.h"
-#elif defined(KRML_VERIFIED_UINT128)
+#elif defined(KRML_VERIFIED_UINT128) && !defined(__clang__)
```
msprotz: I think this one shouldn't be here. 

You still want to enable clang64/windows and verified uint128. This line would make this scenario impossible (I believe).

msprotz:
This looks almost perfect. There's one nit that I've flagged -- once you've fixed it, please go ahead and merge. Thanks for a much appreciated patch!

franziskuskiefer:
This required another small change but should be good now.
Note that `fstar_uint128_gcc64.h` might not be the best name anymore now. It's more like `fstar_uint128_native.h`.

Issue #166, opened 2020-02-05
Doesn't build on ppc64le


clbr:
NSS now includes a part of kremlin, and this causes the NSS build to fail on ppc64le. The following patch let me build NSS successfully. I dont know what Kremlin is, what it does, or how to test it; but being a GCC platform it should be correct.

https://termbin.com/q42p

msprotz:
@franziskuskiefer can I ask you for a sanity check on this one?

@clbr why the restriction on little-endian?

clbr:
The code appeared to only be valid on LE, the other arches enabled were LE only. That, and I only have a LE machine.

clbr:
Oh, you can remove the #include "prcpucfg.h", that's an NSS thing (I used a NSS endian macro at first, but __LITTLE_ENDIAN__ is a GCC one).

franziskuskiefer:
Looks good to me as long as ppc64le has uint128 support :smile: 
I opened #167 with the patch.

franziskuskiefer:
Fixed by #167. Please re-open if it makes any issues. Note that this will need to find its way to NSS before you'll see the NSS build working again.

glandium:
NSS is similarly broken on ppc64 (big-endian), mips64el, s390x, ia64, riscv64, and sparc64.

beurdouche:
@glandium Do we have a CI somewhere that shows us that ?

glandium:
https://buildd.debian.org/status/package.php?p=nss

beurdouche:
Thank you ! I’ll try to have a look at it.

PR #167, opened 2020-02-05
Fix build on ppc64le

franziskuskiefer:
Fixes issue #166


msprotz:
Looks good to me, thanks very much for double-checking Franziskus!

Issue #169, opened 2020-02-21
Build failure


fangyi-zhou:
```
ocamlbuild -I src -I lib -I parser -I kremlib -use-menhir -use-ocamlfind -classic-display -menhir "menhir --infer --explain" Kremlin.native
ocamlfind ocamlc -c -g -annot -bin-annot -strict-sequence -w @1-2-3@8..12@14..21@23..29-30@31..38-39-40-41@43@57 -package ppx_deriving.std -package ppx_deriving_yojson -package zarith -package pprint -package unix -package menhirLib -package sedlex -package process -package fix -package wasm -package visitors.ppx -package visitors.runtime -I src -I parser -I kremlib -I lib -o src/CFlatToWasm.cmo src/CFlatToWasm.ml
+ ocamlfind ocamlc -c -g -annot -bin-annot -strict-sequence -w @1-2-3@8..12@14..21@23..29-30@31..38-39-40-41@43@57 -package ppx_deriving.std -package ppx_deriving_yojson -package zarith -package pprint -package
unix -package menhirLib -package sedlex -package process -package fix -package wasm -package visitors.ppx -package visitors.runtime -I src -I parser -I kremlib -I lib -o src/CFlatToWasm.cmo src/CFlatToWasm.ml
File "src/CFlatToWasm.ml", line 196, characters 18-32:
196 |   [ dummy_phrase (W.Ast.TeeLocal (mk_var (env.n_args + 2)));
                        ^^^^^^^^^^^^^^
Error: Unbound constructor W.Ast.TeeLocal
Command exited with code 2.
Makefile:22: recipe for target 'minimal' failed
make: *** [minimal] Error 10
```

I am using wasm 1.1, maybe that's the problem?

msprotz:
ah, I was unaware that the wasm package had been updated -- do you know what replaced the TeeLocal instruction?

fangyi-zhou:
Is Kremlin supposed to support OCaml >= 4.08.0?
The opam package wasm 1.0 supports < 4.08.0, wasm 1.1 supports >= 4.08.0.
I'm using 4.08.1 so I encountered the issue.

Therefore I don't think this would be an easy fix... Maybe it's necessary to contact wasm maintainer to relax the version requirements

msprotz:
Unfortunately, for various reasons in Everest, we can only work with at most OCaml 4.06.1... so this error flew under our radar. 

[...]

PR #170, opened 2020-02-21
Update OCaml version upper bound

fangyi-zhou:
wasm 1.0 doesn't support OCaml >=4.08.0 and wasm 1.1 has incompatible APIs
See also #169


msprotz:
Thanks!

PR #173, opened 2020-03-04
Fix build on s390x; related to #166

ueno:
Even after #166 got merged, the NSS build is still failing on s390x as mentioned in:
https://bugzilla.mozilla.org/show_bug.cgi?id=1615557

The architecture has [several differences from other arches](https://fedoraproject.org/wiki/Architectures/s390x), but with this change, the build succeeds:
* https://src.fedoraproject.org/rpms/nss/blob/master/f/nss-kremlin-ppc64le.patch
* https://koji.fedoraproject.org/koji/buildinfo?buildID=1463542

### DIFF
--- a/include/kremlin/internal/types.h
+++ b/include/kremlin/internal/types.h
@@ -52,7 +52,8 @@ typedef const char *Prims_string;
 #define IS_MSVC64 (defined(_MSC_VER) && defined(_M_X64) && !defined(__clang__))
 #define HAS_INT128 ((defined(__x86_64__) || defined(__x86_64) || \
                      defined(__aarch64__) || \
-                     (defined(__powerpc64__) && defined(__LITTLE_ENDIAN__))) || \
+                     (defined(__powerpc64__) && defined(__LITTLE_ENDIAN__)) || \
+                     defined(__s390x__)) || \
                     (defined(_MSC_VER) && !defined(_M_X64) && defined(__clang__)))
 
 /* The uint128 type is a special case since we offer several implementations of
--- a/include/kremlin/internal/types.h
+++ b/include/kremlin/internal/types.h
@@ -49,19 +49,20 @@ typedef const char *Prims_string;
 
 /* The great static header headache. */
 
+#define IS_MSVC64 (defined(_MSC_VER) && defined(_M_X64) && !defined(__clang__))
+#define HAS_INT128 ((defined(__x86_64__) || defined(__x86_64) || \
+                     defined(__aarch64__) || \
+                     (defined(__powerpc64__) && defined(__LITTLE_ENDIAN__))) || \
+                    (defined(_MSC_VER) && !defined(_M_X64) && defined(__clang__)))
+
 /* The uint128 type is a special case since we offer several implementations of
  * it, depending on the compiler and whether the user wants the verified
  * implementation or not. */
-#if !defined(KRML_VERIFIED_UINT128) && defined(_MSC_VER) && defined(_M_X64) && \
-    !defined(__clang__)
+#if !defined(KRML_VERIFIED_UINT128) && IS_MSVC64
 #  include <emmintrin.h>
 typedef __m128i FStar_UInt128_uint128;
-#elif !defined(KRML_VERIFIED_UINT128) && !defined(_MSC_VER) && \
-      (defined(__x86_64__) || defined(__x86_64) || defined(__aarch64__) || \
-      (defined(__powerpc64__) && defined(__LITTLE_ENDIAN__)))
+#elif !defined(KRML_VERIFIED_UINT128) && HAS_INT128
 typedef unsigned __int128 FStar_UInt128_uint128;
-#elif !defined(KRML_VERIFIED_UINT128) && defined(_MSC_VER) && defined(__clang__)
-typedef __uint128_t FStar_UInt128_uint128;
 #else
 typedef struct FStar_UInt128_uint128_s {
   uint64_t low;
@@ -75,15 +76,13 @@ typedef FStar_UInt128_uint128 FStar_UInt128_t, uint128_t;
 
 #include "kremlin/lowstar_endianness.h"
 
-/* This one is always included, because it defines C.Endianness functions too. */
-#if !defined(_MSC_VER) || defined(__clang__)
+#if !defined(KRML_VERIFIED_UINT128) && HAS_INT128
 #include "fstar_uint128_gcc64.h"
-#endif
-
-#if !defined(KRML_VERIFIED_UINT128) && defined(_MSC_VER) && !defined(__clang__)
+#if !defined(KRML_VERIFIED_UINT128) && IS_MSVC64
 #include "fstar_uint128_msvc.h"
 #elif defined(KRML_VERIFIED_UINT128)
 #include "FStar_UInt128_Verified.h"
+#include "fstar_uint128_struct_endianness.h"
 #endif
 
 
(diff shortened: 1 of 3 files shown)


include/kremlin/internal/types.h
```diff
@@ -58,7 +58,8 @@ typedef const char *Prims_string;
 typedef __m128i FStar_UInt128_uint128;
 #elif !defined(KRML_VERIFIED_UINT128) && !defined(_MSC_VER) && \
       (defined(__x86_64__) || defined(__x86_64) || defined(__aarch64__) || \
-      (defined(__powerpc64__) && defined(__LITTLE_ENDIAN__)))
```
franziskuskiefer: With this getting more and more complex we should maybe add a custom define for this so we have only one place where this is specified.

beurdouche:
I suggest we do, I suspect these macros are breaking the Firefox build that I am debugging right now, so regrouping them would be useful...

beurdouche:
I think I like that.

ueno:
Thank you for the suggestion; although I haven't tested it yet, it might be something like this?

include/kremlin/internal/types.h
```diff
@@ -75,15 +77,13 @@ typedef FStar_UInt128_uint128 FStar_UInt128_t, uint128_t;
...
+#if !defined(KRML_VERIFIED_UINT128) && HAS_INT128
 #include "fstar_uint128_gcc64.h"
-#endif
-
-#if !defined(KRML_VERIFIED_UINT128) && defined(_MSC_VER) && !defined(__clang__)
+#if !defined(KRML_VERIFIED_UINT128) && IS_MSVC64
 #include "fstar_uint128_msvc.h"
 #elif defined(KRML_VERIFIED_UINT128)
 #include "FStar_UInt128_Verified.h"
+#include "fster_uint128_struct_endianness.h"
```
franziskuskiefer: typo `fstar_uint128_struct_endianness`

include/kremlin/internal/types.h
```diff
@@ -49,19 +49,21 @@ typedef const char *Prims_string;
 
 /* The great static header headache. */
 
+#define IS_MSVC64 (defined(_MSC_VER) && defined(_M_X64) && !defined(__clang__))
+#define HAS_INT128 ((defined(__x86_64__) || defined(__x86_64) || \
+		     defined(__aarch64__) || \
```
franziskuskiefer: Can you adjust indentation here to make this a little easier to read (the `defined` is in that bracket above).

include/kremlin/internal/types.h
ueno: Ugh, fixed ;-(

include/kremlin/internal/types.h
```diff
@@ -49,19 +49,21 @@ typedef const char *Prims_string;
 
 /* The great static header headache. */
 
+#define IS_MSVC64 (defined(_MSC_VER) && defined(_M_X64) && !defined(__clang__))
+#define HAS_INT128 ((defined(__x86_64__) || defined(__x86_64) || \
+                     defined(__aarch64__) || \
+                     (defined(__powerpc64__) && defined(__LITTLE_ENDIAN__)) || \
+                     defined(__s390x__)) || \
+                    (defined(_MSC_VER) && defined(__clang__)))
```
msprotz: could this last check be true in a hypothetical 32-bit build on windows with clang?

msprotz: which would then not have int128

kremlib/c/fstar_uint128_struct_endianness.h
```diff
@@ -0,0 +1,59 @@
+/* Verified uint128 implementation. */
```
msprotz: missing copyright header here

msprotz: also please update the comment: hand-written implementation of endianness-related uint128 functions for the extracted uint128 implementation

msprotz:
this generally looks good -- our CI system has a bug so I'll have to push this to a branch on the fstarlang/kremlin repo, it looks like builds aren't succeeding for external forks

can you sign the CLA?

thanks!!

kremlib/c/fstar_uint128_struct_endianness.h
msprotz: also missing header inclusion guard

include/kremlin/internal/types.h
ueno: I've added a `!defined(_M_X64)` check here; would it be sufficient?

ueno:
Thanks for the review; I've signed the CLA a while ago.

msprotz:
thanks -- for some reason the CLA bot was complaining earlier

@beurdouche is there any chance you could give this a go at Mozilla to make sure it doesn't break any of your platforms? that way we can anticipate rather than break you next time you pull this code

msprotz:
ping @beurdouche or @glandium would this fix the Firefox build on various platforms that Debian packages? I'm happy to merge this, I would just like a sanity check that this fixes the build issue

glandium:
It fixes building NSS on Debian s390x. I haven't checked other architectures.

msprotz:
Thanks Mike. I guess I'll just try to merge this and see what happens, unless you know how to run a candidate build for other Debian-supported platforms (I am no Debian developer so I have no clue).

FlyGoat:
https://buildd.debian.org/status/package.php?p=nss&suite=sid
It fails on mips64el & riscv64 as well, should we add these two architectures to the if defined?

Thanks.

msprotz:
I've pushed a working version of this PR based on @ueno's sketch (there were multiple issues, including a non-terminated #ifdef along with a #define defining some defined(...) which clang helpfully let me know is UB.

I also integrated @FlyGoat's suggestions for the other two platforms. This is on branch protz_feature_detection of kremlin.

@FlyGoat can you let me know if this fixes the build? I've also consolidated the ifdefs so that even if we think the platform doesn't support uint128, we always fallback to the verified (albeit, slow) implementation of uint128 using a pair of uint64. For some reason, this was not done before -- I'm hoping this makes the header more resilient and "always working".

Meanwhile, I'm running some Everest regression to make sure this doesn't break anything either, including in our compilation path with MSVC.

Thanks everyone for the feedback and suggestions, much appreciated.

msprotz:
Everest CI came back green so I'm intending to merge protz_feature_detection at the end of the week.

FlyGoat:
Can confirm it works on MIPS64EL.
Thanks!

msprotz:
Wonderful, thanks for the help debugging!

include/kremlin/internal/types.h
```diff
@@ -49,19 +49,21 @@ typedef const char *Prims_string;
 
 /* The great static header headache. */
 
+#define IS_MSVC64 (defined(_MSC_VER) && defined(_M_X64) && !defined(__clang__))
+#define HAS_INT128 ((defined(__x86_64__) || defined(__x86_64) || \
+                     defined(__aarch64__) || \
+                     (defined(__powerpc64__) && defined(__LITTLE_ENDIAN__)) || \
+                     defined(__s390x__)) || \
+                    (defined(_MSC_VER) && !defined(_M_X64) && defined(__clang__)))
```
glandium, line 57: Note that `__SIZEOF_INT128__` could be used to automatically add support to the platforms that aren't explicitly listed here but have compiler support for __int128.

msprotz: thanks, I've added that too in c3322295890570aeea0f240c0257910a907054de

beurdouche:
This is now in NSS master (should be in 3.53). Thanks all !
https://github.com/nss-dev/nss/commit/184083346d77163ba2b8bd4ac8eb088618d2f36b

PR #181, opened 2020-06-22
Undefine low and high at the end of fstar_uint128_msvc.h

sopium:
Letting them leak outside of the header can break unrelated code: https://github.com/project-everest/hacl-star/issues/316.


msprotz:
This is great! Thanks very much!

PR #191, opened 2020-08-18
Revamping include guards in kremlin

msprotz:
Opening up a pull request to discuss this.

@niekbouman I had to fix up a few things. The tricky bit is that with the new include guards, a dependency loop is starting to cause some trouble. The dependency is as follows:

FStar_UInt_8_16_32_64.h includes types.h which includes fstar_uint128.h which in turn needs functions from FStar_UInt_8_16_32_64.h

the way I broke the loop is to not include fstar_uint128.h when types.h is included from FStar_UInt_8_16_32_64.h -- but this is really not great, as a client header could conceivably #include FStar_UInt_8_16_32_64.h only to find out that it forgot to include types.h *first* in order to get the uint128 definitions into scope

it seems to work based on preliminary testing but I'm not super happy with this; do you have any advice or a suggestion on how to make it more robust?


niekbouman:
Hi Jonathan,
I wouldn't call myself a header-file-expert; the following suggestions stem from some experience and from "plain googling":

The takeaway is that circular header dependencies are to be avoided.
And I agree that your current solution sounds brittle.

The common advice is to avoid header circularities by making use of forward declarations, 
and breaking (existing) circularities by factoring-out certain code-parts into their own header:
https://stackoverflow.com/questions/4816698/avoiding-circular-dependencies-of-header-files
https://stackoverflow.com/questions/4734998/circular-dependency-best-practice/4735091#4735091

The following tutorial article talks about C++ but also applies to C
http://www.cplusplus.com/articles/Gw6AC542/

Hope this helps.

msprotz:
Agreed, circular header dependencies are bad, but this case is slightly more complex than just regular splitting of dependencies to avoid loops. We are extracting kremlib, doing default code-gen that results in #include types.h which itself expects kremlib to be already extracted in order to be able to successfully bring uint128 functions into scope.

The natural way to fix this is to have an extraction mode that says "currently extracting kremlib, uint128 not yet defined, don't try to bring it into scope". Checking whether FStar_UInt_8_16_32_64.h is defined is an ok approximation of that, since this file (with this particular bundling/naming scheme) only appears when extracting kremlib.

So, I merged and pushed what I had before. Thanks for the pointers!

Issue #190, opened 2020-08-14
Consider handling ocaml-ctypes bindings in the presence of undefined C abstract types that are only manipulated through pointers


msprotz:
Hello @victor-dumitrescu,

Son and I were debugging bindings a little further, and we found the interesting use-case below:

```
module Test

module B = LowStar.Buffer
module ST = FStar.HyperStack.ST

open FStar.HyperStack.ST

[@ CAbstractStruct ]
type t = {
  x: UInt32.t;
  y: UInt128.t;
}

type u = B.pointer t

let create (): St u =
  B.malloc FStar.HyperStack.root ({ x = 0ul; y = Int.Cast.Full.uint64_to_uint128 0UL }) 1ul

```

what happens here is that `t` is a "C abstract struct", i.e. it is not defined in the header, meaning that any function in the header will pass *pointers* to it.

```
typedef struct Test_t_s Test_t;

Test_t *Test_create();
```

This is a standard C programming idiom that forces clients of this header to go through the allocation functions exposed in the header, and prevents clients from seeing the internal definition of the struct type. (This is the best you can get in terms of abstraction in C).

Here, the type `t` cannot be described in ocaml-ctype's DSL, because it uses the uint128 type. However, if there was a way to bind the type `t` as an opaque type, then we could happily let ocaml clients manipulate opaque pointers to a struct whose definition has not been exposed by ocaml-ctypes.

Do you know if this is something that's feasible in ocaml-ctypes and whether we could have support for it? That way, we would make sure that i) kremlin's ctypes generation honors C abstract types and ii) Son's code, which uses non-bindable types but only behind a pointer, compiles properly.

Right now, this is what the bindings file contains:

```
open Ctypes
module Bindings(F:Cstubs.FOREIGN) =
  struct
    open F
    let test_create =
      foreign "Test_create" (void @-> (returning (ptr test_t)))
```

is there any way to do

```
let test_t = opaque "Test_t"
```
?

Thanks!

Jonathan

victor-dumitrescu:
Assuming I understood the situation correctly, I think this is something we already do in other places, for example, in EverCrypt.AEAD.fsti

```
[@CAbstractStruct]
val state_s: alg -> Type0
```
gets compiled to `typedef struct EverCrypt_AEAD_state_s_s EverCrypt_AEAD_state_s;` in EverCrypt_AEAD.h and gets this Ctypes binding:
```
[...]

PR #192, opened 2020-08-21
GenCtypes: Bind abstract structs accordingly

victor-dumitrescu:
This PR changes the handling of abstract structs when generating Ctypes bindings. Currently these are treated like any other structs, meaning that:

1. Abstract structs which depend on unsupported types are not bound at all (#190)
2. When bound, their internal structure is exposed.

This PR addresses both these issues.


msprotz:
Looks good, thanks Victor!

PR #193, opened 2020-09-18
add -skip-makefiles: do not generate Makefile.basic, Makefile.include

tahina-pro:
`Makefile.basic` and `Makefile.include` are not always needed, in particular, if KReMLin needs to be called several times in parallel to produce disjoint output. So this pull request adds a new option, `-skip-makefiles`, to not generate them.


msprotz:
Perfect! Nothing to nitpick about, thanks Tahina!

PR #195, opened 2020-09-29
GenCtypes: Improvements to warnings

victor-dumitrescu:
This PR includes:

* "Dropping declaration" warnings now include the name of the declaration, along with the existing message on which type caused it to be dropped
* All warnings are only triggered once
* Minor formatting nit for warnings


msprotz:
That's pretty good, and I like this patch, thanks!

How much would it be to make it something more general where we hash-cons the warnings that were issued, so that for a given pair of a `(loc, warning)`, we only emit it once, with code that's generic and in the `Warn` module rather than just for the ctypes binding generation?

What do you think?

msprotz:
Hi Victor, this patch was quite great! The hash-consing wasn't very much added work, right? Do you think you'd have to finish and merge this?

victor-dumitrescu:
Hi, yes I don't think this would require that much more work. I can't commit to a specific time at the moment, but I'll try to get round to it when I can.

msprotz (approved):
This is great, thanks Victor

PR #197, opened 2020-10-23
Allow no printing in js loader

raphael-proust:
This MR adds a `setMyPrint` function to `loader.js` that can be used to suppress output (or log to a different channel or some other such thing).

A bonus commit fixes some error messages.

One question remains: should the null-logging be the default and the new feature used to allow users to `setMyPrint(console.log)`?

### DIFF
--- a/kremlib/js/loader.js
+++ b/kremlib/js/loader.js
@@ -477,8 +477,14 @@ function link(imports, modules) {
   return fold(i, modules);
 }
 
+function setMyPrint(f) {
+  my_print = f;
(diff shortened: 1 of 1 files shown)


msprotz:
Looks good! I don't think null-logging should be the default for now as errors are quite frequent when one tries to integrate this code so the more debug info, the better. I'll let you sign the CLA then I'll merge. Thanks!

msprotz:
@victor-dumitrescu can we close this one?

PR #198, opened 2020-10-23
Enable the use of intrinsics in wasm code

victor-dumitrescu:
Attempting to extract code which relies on compiler intrinsics to wasm currently fails as the required functions are not present. This PR fixes this by making sure the wasm functions from the fallback F* implementations of intrinsics are always in scope and that their names can be resolved when extracting code that uses them.

### DIFF
--- a/src/CFlatToWasm.ml
+++ b/src/CFlatToWasm.ml
@@ -61,10 +61,20 @@ let name_of_string = W.Utf8.decode
 let string_of_name = W.Ast.string_of_name
 
 let rec find_func env name =
-  try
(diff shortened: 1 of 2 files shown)


msprotz (changes requested):
This looks good; however, can you please add your redirection around line 662? There's already a bunch of similar functions that are being redirected to "alternative" implementations so it'd be good to keep all of those in the same place. Thanks!

msprotz (approved):
All looks good to me, thank you! Please go ahead and merge this PR and the corresponding one on HACL

PR #200, opened 2020-11-14
define KRML_HOST_EPRINTF when building with MSVC

franziskuskiefer:
I finally got around to do the [evecrypt rust bindings for Windows](https://github.com/franziskuskiefer/evercrypt-rust/tree/windows-builds). It turned out that it doesn't actually work unless I'm missing something (I wonder if anyone is using hacl on Windows 😬 ).
This PR fixes the Windows builds for me.


msprotz (approved):
Looks good to me, thanks!

msprotz:
> I wonder if anyone is using hacl on Windows 😬
Yeah... good catch. I suspect people cherry-pick individual algorithms and may even end up defining their custom headers themselves rather than importing all of `kremlin/include` *en masse*

PR #203, opened 2021-03-09
FreeBSD Port

web-vertalo:
Replace make with $(MAKE) and look for bash along PATH.


msprotz:
Thanks! Much appreciateD.

PR #210, opened 2021-06-09
Update a preprocessor condition in testlib.c

Kachoc:
This PR should fix a building issue for Hacl* on ARM.

### DIFF
--- a/kremlib/c/testlib.c
+++ b/kremlib/c/testlib.c
@@ -3,7 +3,8 @@
 
 #include "TestLib.h"
 
-#ifndef _MSC_VER
+#if !defined(_MSC_VER) && \
+    !defined(__aarch64__) && !defined(_M_ARM64) && !defined(__arm__) && !defined(_M_ARM)
 TestLib_cycles TestLib_cpucycles(void) {
   unsigned hi, lo;
   __asm__ __volatile__("rdtsc" : "=a"(lo), "=d"(hi));


franziskuskiefer:
> This PR should fix a building issue for Hacl* on ARM.

Where do you see the ARM issues? The CI is happy. Is there more we should be doing on CI to catch issues like this?

msprotz:
Son is trying to improve HACL*'s test target to be feature-based (tests for 128-bit, tests for 256-bit) rather than platform-based (tests for ARM, etc.).

In the process of doing so, I believe he enabled a 128-bit test on ARM that needed testlib.c and this resulted in the error. I think his PR on the HACL side will provide adequate testing. Bear in mind this this is just for CPU cycle-count code.

@Kachoc the ifdef test is slightly incorrect -- I believe you want to say "if this is GCC *and* x86-64 or x86", because this is really the only situation in which the inline ASM code works

msprotz:
By the way thanks for taking a look @franziskuskiefer -- much appreciated!

protz:
Ok so here's what was happening.
- The "full" kremlib with support for the test helpers was not building on non-x86 platforms.
- Son enabled the build of 128-bit tests and extracted C tests on HACL* for non-x86 platforms. The latter needs the "full" kremlib.

I fixed this PR (7db5eb6) so that it properly addresses the first point. We should be good now. Thanks to all involved!

Kachoc:
Great, thanks !

PR #218, opened 2021-12-22
Grow the WASM memory if the data being passed in is too large to fit …

### LINKED ISSUE #217: The js backend should better handle OOM
The `reserve` function in `loader.js` doesn't check that there is enough memory before returning.
This can result in the following error when accessing the "reserved" memory. 

```
RangeError: offset is out of bounds
```

This was spotted using `hacl-star` with large input.

- consider dynamically growing the memory buffer (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Memory/grow)
- make `reserve` raise a proper OOM error.

msprotz:
…in the memory. This fixes #217

### DIFF
--- a/kremlib/js/loader.js
+++ b/kremlib/js/loader.js
@@ -448,7 +448,15 @@ function propagate(module_name, imports, instance) {
 function reserve(mem, size) {
   let m32 = new Uint32Array(mem.buffer);
   let p = m32[0];
-  m32[0] += size;
+  let curr = mem.byteLength;
+  let want = p + size;
+  if (want > curr) {
+    // The number of WebAssembly pages you want to grow the memory by (each one is 64KiB in size).
+    let need = Math.ceil((want - curr) / (64 * 1024)) + 1;
+    // One extra page for suitable stack space.
+    mem.grow(need);
+  }
+  m32[0] = want;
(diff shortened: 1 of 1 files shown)


msprotz:
@hhugo can you please give this a try and let me know if this fixes your problem?

hhugo:
I don't have time to check this now but it looks very similar to what I hack myself while debugging.

Thinking more about this. It looks like we never reclaim any memory. Is there any reason no to use malloc + free ?

msprotz:
The JS bindings for HACL restore the stack pointer to reclaim the memory here: https://github.com/project-everest/hacl-star/blob/master/bindings/js/api.js#L284

Regarding malloc & free: the kremlin-to-WASM direct compilation pipeline does not depend on any libc and does not implement malloc/free, only stack allocations with a simple bump-pointer allocator. The top-of-the-stack mark is a uint32 at index 0 of the memory, and initially starts with 128 to leave a little space for a simple debug protocol I have been using in the past (see loader.js:400).

msprotz:
The HACL code does not use malloc or free, and therefore, can be safely compiled by the kremlin-to-WASM direct pipeline.

msprotz:
@hhugo can you tell me whether this fixes your issue before I merge? thanks!

hhugo:
It does fix my issue. Thanks

PR #219, opened 2022-01-03
Internal headers

msprotz:
This is a pretty substantial change to the way KReMLin generates code, so I'm opening up this PR as a heads-up.

### Problem statement

I'm going to use concrete module names to help describe the problem and the solution. Consider three modules:
- Curve25519, a module which contains `secret_to_public`, a public-facing function (i.e. exposed in `Curve25519.fsti`) intended to be visible to C clients.
- Curve25519.Internal, a module which contains auxiliary definitions used by `secret_to_public`. Having a separate module is important to enable parallel batch verification and for modularity in general; the module contains a function called `fsquare_times` which is exposed to other F\* modules (via `Curve25519.Internal.fsti`), but is not intended to be exposed to C clients.
- Ed25519, a module that implements the signature algorithm, and in the course of doing so, calls `Curve25519.Internal.fsquare_times`.

Users of KReMLin will typically request that Curve25519 and Curve25519.Internal be grouped in the same C file, using an argument of the form: `-bundle Curve25519=Curve25519.Internal`, meaning: "Curve25519 contains public definitions and remains as-is; everything from Curve25519.Internal is considered "private", meaning it goes in Curve25519.c, but is not exposed in `Curve25519.h`".

This *should* result in a `Curve25519.h` that contains `secret_to_public`, and nothing else. This is what we want: a clean C header that only includes what the user wants to reveal as their public C API.

However, there is an exception: if `Ed25519.c` needs to call `fsquare_times`, then KReMLin will make the definition of `fsquare_times` public; this means KReMLin will generate a prototype for `fsquare_times` in `Curve25519.h`: without it, `Ed25519.c` would not compile.

This latter scenario is unpleasant, since it results in `Curve25519.h` exposing to the C client an internal definition that never should have made it into the C public API.

### Proposed solution

With this PR, KReMLin now generates "internal" headers: definitions that the user did not intend to expose to C clients (according to the user-provided `-bundle` flags), but that are called from other compilation units (i.e. other C files), go into internal headers.

Concretely, KReMLin will generate `internal/Curve25519.h` that contains the prototype for `fsquare_times`.

- `Curve25519.h` does *not* include `internal/Curve25519.h`
- `Curve25519.c` *does* include `internal/Curve25519.h` (rather than repeat prototypes in `Curve25519.c`)
- `Ed25519.h` does *not* include `internal/Curve25519.h`
- `Ed25519.c` *does* include `internal/Curve25519.h`
- `internal/Curve25519.h` *does* include `Curve25519.h`, because some type definitions from `Curve25519.h` may need to be in scope to declare prototypes in `internal/Curve25519.h`

The net effect is that these definitions remain "private" to C clients, who do not include the internal headers. 

### Notes

This situation arises when using `-bundle`, but can also happen without the presence of `-bundle` flags, such as:
- using `friend`
- using `inline_for_extraction`
- using `StackInline`

### Next steps

I have an Everest green with this change.

Screen Shot 2022-01-03 at 9 17 25 AM

I have submitted a draft PR on HACL* with a few minor changes. The effect on `dist/` can be seen quite starkly: https://github.com/project-everest/hacl-star/pull/506/files#diff-825530fa892727d4bda5515ea09990a19b6407f82de16546e42635fe078e0890L43-L52 is the example I used here, but there are many more such chunks that just go away and stop polluting public headers.

I intend to merge this at the end of the week unless someone has concerns.

Thanks.


beurdouche:
This change is pretty nice, thank you!

kremlib/js/loader.js
```diff
@@ -448,7 +448,15 @@ function propagate(module_name, imports, instance) {
 function reserve(mem, size) {
   let m32 = new Uint32Array(mem.buffer);
   let p = m32[0];
-  m32[0] += size;
+  let curr = mem.byteLength;
```
R1kM, line 451: Minor: This change seems to be a standalone commit to fix an unrelated issue

msprotz: yes, I need to rebase without #218, good catch, thanks!

src/Common.ml
```diff
@@ -17,11 +17,11 @@ type flag =
   | Private
       (** An F* private qualifier. *)
   | WipeBody
-      (** Should now be unused. Left there for compatibility with previous ASTs. *)
+      (** DEPRECATED. Left there for compatibility with previous ASTs. *)
```
R1kM, line 20: Would it be worth using OCaml's `deprecated` attribute for this constructors (and the ones below)?

R1kM (approved):

src/Common.ml
msprotz: thanks, I've rebased and fixed this

kremlib/js/loader.js
msprotz: rebase done

tahina-pro (approved):
Thank you Jonathan for this PR. I can confirm that it is compatible with the current way EverParse3D incrementally generates C code: if some module D depends on some other module B whose C code was already generated beforehand, then this PR appears to be compatible with `-library B -bundle D=B`.
Your PR also brings up an interesting opportunity to simplify handling of non-entrypoint validators by declaring them in internal headers instead of resorting to a wrapper. I am going to open an issue on EverParse to suggest such simplifications.

msprotz:
Thanks for the review!

PR #220, opened 2022-01-03
Mark projectors and discriminators as private by default

msprotz:
Meaning their prototypes will be eliminated from public headers, or will go into internal headers (unlikely) rather than show up in the public C API.

This is another big code-quality improvement, to be reviewed after #219

### DIFF
--- a/src/Simplify.ml
+++ b/src/Simplify.ml
@@ -409,6 +409,17 @@ let remove_uu = object (self)
       ELet (b, e1, e2)
 end
 
+let remove_proj_is = object
+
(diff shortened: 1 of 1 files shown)


msprotz:
hi Aymeric, I thought you might this PR enjoyable to review, too: it's another "nanopass" that, this time, marks any projector or discriminator as private so that the rest of the code (which you looked at recently) can eliminate them (if unreachable), or at least hide them behind a .h file... thanks!

R1kM (approved):

PR #221, opened 2022-01-18
Support null and is_null Steel functions.

R1kM:
@cmovcc reported failures when attempting to compile C code extracted from Steel. In particular, references to Steel_Reference_null and Steel_Reference_is_null were undefined.

This PR adds support for these two functions, similarly to existing support for LowStar.Buffer.null and LowStar.Buffer.is_null.

### DIFF
--- a/src/Builtin.ml
+++ b/src/Builtin.ml
@@ -199,6 +199,12 @@ let dyn: file =
       with_type void_star (ECast (with_type (TBound 0) (EBound 0), void_star)))
   ]
 
+let steel_reference : file =
+  "Steel_Reference", [
+    mk_val [ "Steel"; "Reference" ] "is_null" (TArrow (TBuf (TAny, false), TBool));
+    mk_val [ "Steel"; "Reference" ] "null" (TArrow (TAny, TBuf (TAny, false)));
+  ]
+
 let lowstar_monotonic_buffer: file =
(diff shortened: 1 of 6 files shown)


msprotz (approved):
This looks good, thanks for the PR. I would suggest you add a very simple test for this. Just add a new file in tests/ that makes use of these two functions, e.g.

```
let main () ... =
  let x = Steel.null in
  if Steel.is_null x then 
    0l
  else
    1l
```

(sketch)

R1kM:
Good point, thanks for the review. I added one test for null, and one for not_null.

msprotz:
All good for me. You can merge once it's green. Thanks so much!

R1kM:
@msprotz I had to tweak the tests Makefile to work around a known issue with Steel extraction, namely, FStar.NMST and similar files not extracting due to reification. Please let me know if you think of a better way to do this.

msprotz:
Thanks for the fix. It's unfortunate that anyone who wishes to extract Steel to C needs to be aware of this quirk.

One option would be to insert try-catch in FStar.Extraction.Kremlin.fs like we do for some other things that already do not extract (e.g. polymorphic assumes), and simply emit a warning saying that a given definition could not be extracted. That way, Steel extraction would work by default, rather than failing by default.

PR #223, opened 2022-01-25
Extend Builtins with Steel null and is_null

R1kM:
As discussed today, Steel.Reference is not extracted by KreMLin, hence, the null and is_null functions that are not rewritten by FStar.Extraction.Kremlin.fs need to be provided as builtins to KreMLin.

This PR also marks the warning 2 as an error on the Steel tests to detect possible regressions.


msprotz:
Were you able to confirm the warning 2 is gone? If so, please remove steel_reference from the list here: https://github.com/FStarLang/kremlin/pull/223/files#diff-8a10cfa26a87bbd37d3c4409e456d459faa850823bca2c9220008a17a24d79e7L307 then go ahead and merge! Thanks for catching this!

R1kM:
The warning 2 is gone, but only if steel_reference is in the list of builtins so that Steel.Reference.null and is_null have a model.

R1kM:
As discussed on Slack, I removed steel_reference from the "hand_written" list since it's unneeded there.
Thanks!

PR #228, opened 2022-02-14
Native extraction of Steel.ST loop combinators

aseemr:
I added code following similar combinators in `C.Loops`. It works on the Zeta code where we are using these.


msprotz:
Looks great, any chance you can add brief test for this like @R1kM did earlier? He even added the build logic the filter-out the right krml files from the build.

aseemr:
Sorry I pushed it by mistake, will add the test case with Aymeric today.

(Through some series of git commands, I ended up merging this branch to master locally, and then did a `git push` thinking I was pushing my branch rather than `master`, and it just went through.)

msprotz:
Yeah that's my bad I disabled branch protection for administrators to urgently revert a faulty commit, then forgot to enable it back.

msprotz:
The setting should be good now :-)

PR #229, opened 2022-02-14
Use POSIX-friendly syntax

concatime:
Hi, with this patch, I was able to compile Kremlin on a POSIX-friendly environment, which means:
1. replace `which` by the POSIX `command -v`
2. replace `bash` by POSIX `sh`

This helps a little bit more the build process by removing two dependencies.


msprotz:
Thanks for your contribution! Could you tell us a little bit why this is important? I buy the changes related to `which`, but why get rid of `bash`? This complicates the behavior in that we can't really rely on bash-like wildcard expansion which in turn means that, if we want to "do things right", we have to use `$(wildcard )` a lot in Makefiles. Also, if you're building kremlin, surely you must have bash installed somewhere, no?

concatime:
On my local build, bash was not required, so I wrongly thought that I could simply swap it. That being said, from the CI (even tho I cannot access the log because it requries that I log-in to a Microsoft account, seriously?), it seems important. So yeah, you're right. I restored the `SHELL` to bash, and will install it on my machine.

PR #230, opened 2022-02-14
New syntax for adding #includes in C files only

### LINKED ISSUE #182: -add-early-include and -add-include behave the same
The two options
```
-add-early-include      prepend #include the-argument to every generated file, before kremlib.h
-add-include            prepend #include the-argument to every generated file, after the #define __FOO_H
```
behave both like the former, resulting in the following kind of output 
```
#include "manually_included.h"
#include "kremlib.h"
#ifndef __FOO_H
#define __FOO_H
...
```

msprotz:
This is useful to limit name collisions by not leaking un-namespaced function names to clients of a given module.

(Context: this came up in the context of Curve64, where Vale's un-namespaced `fadd` conflicts with `fadd` from the math library, making it impossible to write a client that uses both Curve64 and the original `fadd`.)

This fixes #182, also.

### DIFF
--- a/src/Kremlin.ml
+++ b/src/Kremlin.ml
@@ -172,9 +172,16 @@ Supported options:|}
   let used_drop = ref false in
   let prepend r = fun s -> r := s :: !r in
   let parse_include s =
+    let open Options in
     match String.split_on_char ':' s with
-    | [ h; i ] -> Some h, i
-    | [ i ] -> None, i
+    | [ h; i ] ->
+        begin match Filename.chop_suffix_opt ~suffix:".c" h with
+        | Some h -> 
+            COnly h, i
+        | None ->
+            HeaderOnly h, i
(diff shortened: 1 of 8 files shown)


src/Kremlin.ml
```diff
@@ -401,7 +408,7 @@ Supported options:|}
 
     (* Self-help. *)
     if Options.debug "force-c" then begin
-      Options.add_include := (None, "\"kremlin/internal/wasmsupport.h\"") :: !Options.add_include;
+      Options.add_include := (All, "\"kremlin/internal/wasmsupport.h\"") :: !Options.add_include;
```
R1kM, line 411: Why do you need to include wasmsupport when using the force-c option?

R1kM (approved):
Looks good to me, thanks for the change!

PR #231, opened 2022-02-14
Make warning 2 fatal for tests.

msprotz:
@R1kM I was prompted to do this following your recent PRs. Warning 2 should never happen anymore now that we have suitable -library, -static-header and combination of those flags. There was one test that I'm leaving here for posterity but it is a very incorrect use-case and I do not condone this sort of stuff anymore.


R1kM:
Thanks! I'm wondering whether it would make sense to make warning 2 fatal by default in KreMLin, instead of just in tests. In my recent experience, the presence of a warning 2 was a bit lost in the output log, and the fact that it was more problematic than other warnings was lost on me.
Are you aware of any existing code that relies on warning 2 not being fatal?

msprotz:
I had an Everest green. @tahina-pro might needs a heads-up

tahina-pro (approved):

PR #233, opened 2022-02-25
Remove side effect in driver mode

pnmadelaine:
The use of `--cache_checked_modules` in the call to fstar triggers generation of `.checked.lax` files in all the included paths.
It appears unnecessary and can create issues, typically if a read-only path is included.


msprotz:
Thanks Paul-Nicolas! The ICFP deadline has passed so I can merge this and keep an eye on potential regressions, now.

PR #235, opened 2022-03-09
Rename KreMLin into KaRaMeL in book/

R1kM:
This PR propagates the project renaming to book/ and hence to the Low* tutorial.
Minor: It also relaxes the upper bound on OCaml, which does not seem required anymore according to opam.

Note: The everest script and ecosystem currently still uses variables KREMLIN_HOME as well as the kremlin target, hence I did not modify those (yet)


msprotz:
Thanks Aymeric for your help!

PR #250, opened 2022-04-13
Native extraction of Steel loops

R1kM:
Extends #228 with extraction of other Steel loops.
The code path and signatures are exactly the same as for the Steel.ST loops.


msprotz (approved):
Looks good to me, but can you please add a test before merging? Thanks!

R1kM:
Thanks Jonathan for the comments. I added tests in SteelLoops and confirmed that the outputs looked like expected.
I also cleaned up slightly the test Makefile to be less agressive in the number of Steel files that need to be filtered.

PR #252, opened 2022-04-14
make cases for lowstar_endianness consistent

franziskuskiefer:
I don't mind which version to use but the warnings about this are annoying, let's decide on a casing.

### DIFF
--- a/krmllib/dist/generic/fstar_uint128_gcc64.h
+++ b/krmllib/dist/generic/fstar_uint128_gcc64.h
@@ -25,7 +25,7 @@
 
 #include "FStar_UInt128.h"
 #include "FStar_UInt_8_16_32_64.h"
-#include "LowStar_Endianness.h"
+#include "lowstar_endianness.h"
 
 /* GCC + using native unsigned __int128 support */
 
--- a/krmllib/dist/minimal/Makefile.include
+++ b/krmllib/dist/minimal/Makefile.include
@@ -2,4 +2,4 @@ USER_TARGET=libkrmllib.a
 USER_CFLAGS=
 USER_C_FILES=fstar_uint128.c
 ALL_C_FILES=
-ALL_H_FILES=FStar_UInt128.h FStar_UInt_8_16_32_64.h LowStar_Endianness.h
+ALL_H_FILES=FStar_UInt128.h FStar_UInt_8_16_32_64.h lowstar_endianness.h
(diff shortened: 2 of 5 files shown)


msprotz:
PascalCase = kremlin-generated file
snake_case = hand-written

there are two such files: the snake_case one is in include/kremlin, the PascalCase one is in dist/generic/*

The correct fix is probably to rename the snake_case one to something more meaningful, such as `lowstar_endianness_builtins.h` to indicate that it contains hand-written prototypes that "fill out" the endianness functions on a per-platform basis.

The auto-generated one contains function prototypes to be implemented by hand and allows making sure that the implementation of e.g. `load128_le` (in `kremlib/c/fstar_uint128_gcc64.h`) is correct with regards to what is claimed in the F* file (as extracted in `kremlib/dist/generic/LowStar_Endianness.h`).

Hope this makes sense!

krmllib/dist/generic/fstar_uint128_gcc64.h
```diff
@@ -25,7 +25,7 @@
 
 #include "FStar_UInt128.h"
 #include "FStar_UInt_8_16_32_64.h"
-#include "lowstar_endianness.h"
```
msprotz: this one seems wrong

krmllib/dist/minimal/Makefile.include
```diff
@@ -2,4 +2,4 @@ USER_TARGET=libkrmllib.a
 USER_CFLAGS=
 USER_C_FILES=fstar_uint128.c
 ALL_C_FILES=
-ALL_H_FILES=FStar_UInt128.h FStar_UInt_8_16_32_64.h lowstar_endianness.h
```
msprotz: same here should be the builtin one

krmllib/dist/minimal/fstar_uint128_gcc64.h
```diff
@@ -25,7 +25,7 @@
 
 #include "FStar_UInt128.h"
 #include "FStar_UInt_8_16_32_64.h"
-#include "lowstar_endianness.h"
```
msprotz: same

msprotz (commented):
This is good, thanks. The files you modified are for a large part auto-generated or copied from another directory. Can you modify the source files in krmllib/c/{fstar_uint128*.h} rather? Then I can take care of refreshing dist after that and check that it's alright. Thanks!

Issue #247, opened 2022-03-30
Monomorphizations leak into interfaces and lead to incorrect code in combination with CAbstractStruct


nikswamy:
Here's a repro:

```
------Abstract.fsti----------
module Abstract

val t (a:Type0) : Type0

val make (#a:Type0) (x:a) : t a
val read (#a:Type0) (x:t a) : a
------Abstract.fst----------
module Abstract

type t a = (bool & a)

let make #a x = (false, x)
let read #a t = snd t
------Parameter.fsti----------
module Parameter

[@@CAbstractStruct]
type param = {
  param_one: bool;
  param_two: bool
}
------Main.fsti----------
module Main
open FStar.HyperStack.ST
module B = LowStar.Buffer

[@@CAbstractStruct]
val handle : Type0

val init (_:unit)
  : ST (B.pointer handle)
       (requires fun _ -> True)
       (ensures fun _ _ _ -> True)
------Main.fst----------
module Main
module HS = FStar.HyperStack
module B = LowStar.Buffer

[@@CAbstractStruct]
noeq
type handle = {
  something: bool;
  another: Abstract.t Parameter.param;
}

let init (_:unit) =
  let h = {
    something = false;
    another = Abstract.make Parameter.({ param_one = true; param_two = false })
  } in
  B.malloc HS.root h 1ul
```

When using krml version 70084646 and `krml -bundle Main=*` leads to

```
#ifndef __Main_H
#define __Main_H

#include "krmllib.h"


typedef struct Parameter_param_s Parameter_param;

typedef struct K___bool_Parameter_param_s
{
  bool fst;
  Parameter_param snd;
}
K___bool_Parameter_param;

typedef struct Main_handle_s Main_handle;

Main_handle *Main_init();


#define __Main_H_DEFINED
#endif
```

With the error

```
✘ [CC,./Main.c] (use -verbose to see the output)
In file included from ./Main.c:8:
./Main.h:21:19: error: field ‘snd’ has incomplete type
   21 |   Parameter_param snd;
```

Note that `Parameter_param` is an alias for an abstract struct but is used in a field of a struct ` K___bool_Parameter_param_s`.

If you remove the `CAbstractStruct` annotation on `Parameter.param` you get:

```
typedef struct Parameter_param_s
{
  bool param_one;
  bool param_two;
}
Parameter_param;

typedef struct K___bool_Parameter_param_s
{
  bool fst;
  Parameter_param snd;
}
K___bool_Parameter_param;

[...]

PR #256, opened 2022-04-26
Fixup the interaction of CAbstractStruct, abstract types, and polymorphic equalities

### LINKED ISSUE #245: krml crashes in Monomorphization.ml
I'm using 70084646b6f5cf01d58a2ed1c0e69d34829ba480

Here's a repro:

Abstract.fsti:

```
module Abstract

[@@CAbstractStruct]
val t : eqtype

val mk_t (_:unit) : t
```

Test.fsti:

```
module Test

val nothing : unit -> bool
```

Test.fst:

```
module Test
open Abstract

type pair = {
  fst : t;
  snd : bool
}


let test (x y:pair) : bool = x = y

let nothing() =
  let x = mk_t () in
  test ({fst=x; snd=true}) ({fst=x;snd=false})
```

I observe the following:

```
krml -bundle Test=Abstract prims.krml FStar_Pervasives_Native.krml FStar_Pervasives.krml Abstract.krml Test.krml
✔ [Monomorphization] ⏱️ 9ms
Fatal error: exception File "src/Monomorphization.ml", line 502, characters 16-22: Assertion failed
```

cc @tahina-pro @arvinda

msprotz:
This fixes #245

Note: the original test was ill-formed, in that it was declaring a C abstract struct but yet declaring a function that had the ability to *construct* values of the struct type without a pointer.

I had to add some custom hand-written files to make the test work -- I assume this is representative of your scenario.

Please confirm this solves the problem you were encountering, thanks!

### DIFF
--- a/src/Inlining.ml
+++ b/src/Inlining.ml
@@ -210,6 +210,7 @@ let cross_call_analysis files =
    * be safely marked as private. The invariant is not established yet. *)
   let private_or_internal = Hashtbl.create 41 in
   let safely_inline = Hashtbl.create 41 in
+  let c_abstract_structs = Hashtbl.create 41 in
 
   (* Constants that will end up being mutable in Wasm because of the compilation
    * scheme of constants as little-endian encoded pre-laid out byte literals
@@ -228,6 +229,9 @@ let cross_call_analysis files =
       if List.mem Private flags && not (Helpers.is_static_header name) then
         (* -static-header takes precedence over private, see CStarToC11.ml *)
         Hashtbl.add private_or_internal name T.Private;
+
+      if List.mem AbstractStruct flags then
+        Hashtbl.add c_abstract_structs name ();
 
       if List.mem Inline flags then
         Hashtbl.add safely_inline name ();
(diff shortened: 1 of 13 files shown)


nikswamy:
#245 was synthetic test case on which I observed krml crashing. If this fixes the crash, then that's good.

Does this PR also address #247? that's the case that I'm really interested in and is more representative of the issue I was observing in my full scenario.

msprotz:
#247 is coming up next -- my hunch was that I had at least to clean this up before tackling the core visibility issue

nikswamy:
I tried out this branch on my sample code and it indeed produces a smaller, better looking interface.

However, I did also notice that the issue I reported in #245 is actually also representative of my application. I have declared a type `t` as CAbstractStruct. And then have a function in my external API whose signature is like `t -> unit`. And krml extracts this to something like this header file:

```
typedef struct t_s t

void f(t x)
```

Even though `f` receives a `t` argument rather than a `t *` argument.

Is there a way for krml to detect this situation and emit a warning such as "Type t is declared as a CAbstractStruct but its values are exchanged on the public API without a pointer indirection" ... or something like that?

As it stands now, because of this misuse of CAbstractStruct, it is possible for F* + krml to produce code without any warnings that will fail to compile in C.

msprotz:
Correct, this has always been an attribute with no actual check behind the scenes that it is being used properly. I agree that it would be nice for krml to warn about this, but this should be a separate issue and discussion. In the meanwhile, I take it that this solves the crash and I'll merge it as soon as I get an Everest green. Thanks for checking!

PR #261, opened 2022-04-29
First attempt at accurate import declarations for WASM modules

msprotz:
Previously, every module was requesting as imports the entirety of the definitions that has previously been generated. This was sound, but inefficient.

Now, we collect the required "external" definitions in a first phase, and only generate import declarations for those external definitions (function or global) that we need.

### DIFF
--- a/src/CFlatToWasm.ml
+++ b/src/CFlatToWasm.ml
@@ -9,6 +9,63 @@ module W = Wasm
 module K = Constant
 
 module StringMap = Map.Make(String)
+module StringSet = Set.Make(String)
+
+(******************************************************************************)
+(* Silly helpers                                                              *)
+(******************************************************************************)
+
+(** We don't make any effort (yet) to keep track of positions even though Wasm
+ * really wants us to. *)
+let dummy_phrase what =
+  W.Source.(what @@ no_region)
+
+let name_of_string = W.Utf8.decode
+let string_of_name = W.Ast.string_of_name
+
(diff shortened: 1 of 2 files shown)


src/CFlatToWasm.ml
```diff
@@ -9,62 +9,143 @@ module W = Wasm
...
-let empty = {
-  funcs = StringMap.empty;
-  globals = StringMap.empty;
-  n_args = 0;
+let empty () = {
+  funcs = Hashtbl.create 41;
+
+  globals = Hashtbl.create 41;
+
   strings = Hashtbl.create 41;
```
msprotz, line 120: note that this was sharing the hash table across all modules, meaning each data segment was accumulating the previous modules' data

denismerigoux: Why 41 for the `Hashtbl` initial size ?

denismerigoux (approved):
Looks reasonable to me on the surface  but I was not familiar at all with the previous codebase for generating imports so difficult for me to catch a potential subtle bug.

src/CFlatToWasm.ml
msprotz: https://stackoverflow.com/a/16724713/873567

"A prime number is good because it tends to spread values throughout the table."

Don't know if it makes much difference in practice, but it's a habit I have had for a while...

msprotz:
Screen Shot 2022-05-12 at 1 29 13 PM


Everest green

PR #262, opened 2022-05-02
Steel extraction tests: filter out some F* standard library files

tahina-pro:
This PR uses F*'s `--extract` option to exclude some F* standard library modules from extraction, rather than filtering them out from the Makefile.

This is a stopgap change to unblock Karamel's CI until a more long-term solution is adopted in the F* standard library to never extract those files at the first place.

(A potential solution to the latter problem, as discussed with @nikswamy , @aseemr and @protz , could be to tag entire F* standard library modules with `noextract` , but this needs a change in the F* syntax, which does not currently support module-level attributes.)

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -66,9 +66,10 @@ everything: all wasm
 
 ifndef MAKE_RESTARTS
 ifndef NODEPEND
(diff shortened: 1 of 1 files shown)


msprotz (approved):
Thanks Tahina!

PR #267, opened 2022-05-31
Compat changes for wasm 2.0.0

msprotz:
To be merged after https://github.com/ocaml/opam-repository/pull/21446 gets merged.


msprotz:
I added a commit that now allows returning I64's across the boundary as a multi-value made up of two I32s. This is required to expose Hacl_Bignum64_add for @danwallach

src/AstToCFlat.ml
```diff
@@ -899,32 +899,57 @@ and mk_expr (env: env) (locals: locals) (e: expr): locals * CF.expr =
 
 (* See digression for [dup32] in CFlatToWasm *)
 let scratch_locals =
-  [ I64; I64; I32; I32 ]
```
msprotz, line 902: the top-of-the-stack pointer is now always saved in the fifth local

msprotz: (as opposed to being left on the stack, which used to require contortions to swap operands at the top of the stack)

src/AstToCFlat.ml
```diff
@@ -899,32 +899,57 @@ and mk_expr (env: env) (locals: locals) (e: expr): locals * CF.expr =
...
+  [ I64; I64; I32; I32; I32 ]
 
 let msg_i64 = format_of_string "Abort: %s was meant to be hand-written and \
   provided at link-time, but contains an I64 and therefore cannot be called from \
   WASM."
 
-let mk_decl env (d: decl): env * CF.decl option =
+let mk_wrapper orig_name n_args locals =
```
msprotz, line 908: the strategy for returning 64-bit integers to javascript is as follows:
- if the function `f` is part of the public API
- and returns a uint64
then we generate a _packed variant that calls `f`, then decomposes the uint64 return value as a pair of uint32s

since krml does not take advantage of multi-value returns (it should! but these weren't supported when I wrote the backend), a special instruction here allows converting a 64-bit integer into two 32-bit integers...

in the future, it'd be good to add general support for multi-values, and then simply call WasmSupport.u64tou32s of type `uint64 -> uint32 & uint32`

src/CFlatToWasm.ml
```diff
@@ -981,16 +1001,14 @@ let mk_func env { args; locals; body; name; ret; _ } =
     in
     (if name <> "WasmSupport_align_64" then Debug.mk env debug_enter else []) @
     read_highwater @
+    [ dummy_phrase (W.Ast.LocalSet (mk_var (env.n_args + 4))) ] @
```
msprotz, line 1004: here, instead of leaving the top-of-the-(memory)-stack pointer on top of the (operand) stack, we save it in the reserved local

denismerigoux (approved):
Looks good to me, even though I didn't have a good knowledge of that part of the codebase.

W95Psp (approved):
Don't know much about Karamel's extraction to WASM, but this seems good to me :)

PR #271, opened 2022-07-01
Improve the handling of type abbreviations and type monomorphizations

msprotz:
Previously, type abbreviations of the form `type foo = t x` would merely be treated as a hint that the krml-generated monomorphization of polymorphic type `t` applied to its argument `x` ought to be named `foo`. Notably, this had no influence on the placement of monomorphized type `foo`: type abbreviations were substituted away early on, and monomorphized types were inserted at the site of the first use.

This PR changes the behavior as follows:
- type abbreviations are retained in the AST all the way to type monomorphization
- upon encountering a type abbreviation of the form above, type monomorphization immediately inserts the monomorphized instance with the suitable name (in the absence of a type abbreviation, a name is picked automatically and the monomorphized type gets inserted at the first use-site)
- further type abbreviations for `t x`, a.k.a. `foo` are retained (they were previously discarded)
- further uses of `t x` always compile to `foo`, i.e. the first user-provided name found in the topological order of declarations.

Example:
```
A.fsti
let foo = t x

B.fst
let bar = t x
let f (y: bar) = ...
```

Before:
```
A.h
<empty>

B.h
typedef ... foo;
void f(foo y) ...
```

Now:
```
A.h
typedef ... foo; // NOTE: new placement

B.h
typedef foo bar; // NOTE: second type abbreviation retained
void f(foo y) ... // NOTE: doesn't pick `bar` here, no change for this behavior
```


msprotz:
Started a regression for this under Everest branch `protz_ci_monomorphization`

msprotz:
The graph traversal wasn't right for the tuple case and I was failing to insert a gray node which would then results in inconsistencies as exercised by miTLS. This should fix things now and I'll wait for an Everest green before merging.

msprotz:
Screen Shot 2022-07-25 at 9 09 53 AM

This is green, so I'm going ahead and merging this.

Mostly a ton of fixes for miTLS which was relying on auto-generated type names for monomorphized types, which is generally a bad idea.

PR #272, opened 2022-07-18
Disabling indexed effects reification extraction test

aseemr:
Cf. F* #2641.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -26,7 +26,7 @@ ifneq ($(RECENT_GCC),"yes")
 endif
 
 # See F* bug 2641 and associated PR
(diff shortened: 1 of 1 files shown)


test/Makefile
```diff
@@ -25,6 +25,9 @@ ifneq ($(RECENT_GCC),"yes")
   FILES 	:= $(filter-out Debug.test,$(FILES))
 endif
 
+# See F* bug 2641 and associated PR
+FILES := $(filter-out Layered.test,$(FILES))
```
msprotz, line 29: Why not simply add this test to the list of broken tests 12 lines above? I will be fixing this in #271 since I noticed this while merging master in.

PR #274, opened 2022-07-26
Automatically ignore unused local variables

msprotz:
krml performs a def-use analysis to detect unused variables; if they are trivially readonly, these are just skipped from the code-gen; if they are not obviously readonly (e.g. function call), then the current behavior is to leave an unused variable in the generated code to preserve potential side-effects

an alternate strategy is to cast to `(void)` the definition without introducing a generated variable, so as to silence C compiler warnings, which this PR does

### DIFF
--- a/src/AstToCStar.ml
+++ b/src/AstToCStar.ml
@@ -274,6 +274,12 @@ let rec mk_expr env in_stmt e =
   | EAddrOf e ->
       CStar.AddrOf (mk_expr env e)
 
+  | EIgnore e ->
+      (* An ignore node ended up in expression position. Rather than have a
+         sketchy macro, we just drop the ignore here and make it a "best-effort"
+         thing. *)
+      mk_expr env e
+
   | _ ->
       Warn.maybe_fatal_error (KPrint.bsprintf "%a" Loc.ploc env.location, NotLowStar e);
       CStar.Any
--- a/src/Helpers.ml
+++ b/src/Helpers.ml
@@ -416,8 +416,11 @@ let rec nest_in_return_pos i typ f e =
         t, nest_in_return_pos i typ f e
       ) branches in
       { node = ESwitch (e, branches); typ }
-  | EMatch _ ->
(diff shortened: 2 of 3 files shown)


msprotz:
https://github.com/project-everest/hacl-star/pull/582 tests this patch on a large-scale, and I'm not sure inserting `(void)` casts is even worth it. Would it be more idiomatic in C to just not even insert `(void)`?

msprotz:
@franziskuskiefer would love to have your thoughts

franziskuskiefer:
I think it depends what the best solution here is 😄 
I personally like `(void)`ing unused return values because it makes it explicit and allows us to error on unused variables rather than ignoring the warnings. So I'd vote for adding `(void)`.
But it of course introduces another type of clutter and complexity and many people don't care about unused variables (the compiler will most likely take care of them anyway).

msprotz:
Looking at this again, my take is that if we want the HACL code to *not* look weird, we'll have to do `(void)` insertion along with a manual review to make sure we never insert `(void)f(x);` where `f` is pure in C. So, basically, krml patch + manual touch-ups in HACL to generate higher-quality code.

And then, if users are dissatisfied with the insertion of (void), they can always rework their code to avoid returning a value that ends up being ignored.

karthikbhargavan:
How about we introduce a new function of the form: 
```
let ignore_st (x:'a) : 
     ST unit (requires (fun _ -> True)) (ensures fun h0 _ h1 -> h0 == h1) 
     = ()
```
KaRaMeL could specially translate calls to `ignore_st` to `(void) x`
Now, either the programmer can inject `ignore_st` themselves or KaRaMeL can do it in some cases.

msprotz:
There is a function like this in LowStar.Ignore... not sure how it's compiled right now?

msprotz:
I remember Tahina ran into the exact same problem and I wrote this for him ages ago

msprotz:
Fixed in #384, #385

PR #275, opened 2022-07-27
One more function that we know can be safely removed

msprotz:


msprotz:
Super low hanging fruit found while looking at HACL* code quality.

PR #278, opened 2022-08-11
Special array index to replace b[...] with *b

tahina-pro:
Access to a F* (or Steel) reference `r` is currently extracted to C as `r[0]`, both for reading and writing. Currently, the only way to produce `*r` instead has been to use the special operators in `LowStar.BufferOps`, but this only works with Low* buffers.

**This PR** proposes an alternative way: I introduce a new constant, `C._zero_for_deref`, defined as 0, and I change C code generation from `r[C._zero_for_deref]` to `*r`. This avoids adding a new node in the Karamel source AST.

I have an Everest green (project-everest/everest@35b47bb1ac32eb01f6f507dc65d00ebe0e5609aa) with a companion F* commit (FStarLang/FStar@984a531b50b07794d4edd91bc47ccaaa13546796) that enacts such code generation improvement for F* and Steel references, for which I'll open a F* PR shortly.

### DIFF
--- a/src/CStarToC11.ml
+++ b/src/CStarToC11.ml
@@ -41,6 +41,8 @@ let builtin_names =
     ["C"], "exit";
     ["C"], "fflush";
     ["C"], "clock";
+    (* Special array index to turn b[0] into *b (cf. PR #278) *)
+    ["C"], "_zero_for_deref";
     (* Hand-written type definition parameterized over KRML_VERIFIED_UINT128 *)
     ["FStar"; "UInt128"], "uint128";
     (* Might appear twice otherwise, which is not C89-compatible. Defined by
--- a/src/CStarToC11.ml
+++ b/src/CStarToC11.ml
@@ -846,14 +846,24 @@ and mk_stmts m stmts: C.stmt list =
 
 
 and mk_index m (e1: expr) (e2: expr): C.expr =
-  match mk_expr m e2 with
-  | Cast (_, (Constant _ as c)) ->
-      Index (mk_expr m e1, c)
+  match e2 with
+  | Qualified (["C"], "_zero_for_deref") ->
+      mk_deref m e1
   | _ ->
-      Index (mk_expr m e1, mk_expr m e2)
+    begin match mk_expr m e2 with
+    | Cast (_, (Constant _ as c)) ->
+        Index (mk_expr m e1, c)
+    | e2' ->
+        Index (mk_expr m e1, e2')
+    end
 
 and mk_deref m (e: expr) : C.expr =
-  Deref (mk_expr m e)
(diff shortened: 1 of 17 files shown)


msprotz:
Looks good. I have only one remark: you might find it hard to eliminate code-generation for this specific value, since the dependency analysis (in the current state of your patch) is not aware that there is a special treatment for it down the line.

Have you thought about that?

tahina-pro:
> Looks good. I have only one remark: you might find it hard to eliminate code-generation for this specific value, since the dependency analysis (in the current state of your patch) is not aware that there is a special treatment for it down the line.

Thanks for pointing this out. Yes, I do observe that in my EverParse experiments: a definition of `_zero_for_deref` is inserted somewhere in my bundles. (This also seems to explain why `_zero_for_deref` appears in the generated `dist/generic/C.h`)
I am not aware of any mechanism within Karamel to declare a list of symbols with special treatment for which no C definition should be produced. If there is no such mechanism, then the only way I can see could be to add an attribute to be conveyed down to Karamel and not produce any C code for definitions tagged with that attribute.
Ideally, I would have reused `noextract_to "krml"`, but then the definition already disappears from the `.krml` file and Karamel complains when it is used. That's why we seem to need another separate attribute.
What do you think?

msprotz:
I don't think you need a new attribute.

msprotz:
Thanks, this looks good. I'm not sure your test does very much in the current state of things -- nothing prevents a miscompilation that generates `foo[0]` instead of `*foo`.

I suggest you add a `count-derefs` target (look at this for an example: https://github.com/FStarLang/karamel/blob/master/test/Makefile#L183) that counts the occurrences of `*` and makes sure it breaks if more or less dereferences are introduced in the generated C.

src/CStarToC11.ml
```diff
@@ -921,14 +923,24 @@ and mk_stmts m stmts: C.stmt list =
...
+        Index (mk_expr m e1, c)
+    | e2' ->
+        Index (mk_expr m e1, e2')
+    end
 
 and mk_deref m (e: expr) : C.expr =
-  Deref (mk_expr m e)
+  match mk_expr m e with
+  | Address e' ->
+      (* *&expr is equivalent to expr *)
```
msprotz, line 940: according to the standard, "It is
always true that if E is a function designator or an lvalue that is a valid operand of the unary &
operator, *&E is a function designator or an lvalue equal to E." so looks like we're good here

msprotz (commented):
Looks great thanks Tahina!

msprotz (approved):

tahina-pro:
Thanks a lot Jonathan! I can now proceed with the F* companion PR.

PR #280, opened 2022-08-23
Enable macro-based loop unrolling

msprotz:


include/krml/internal/target.h
```diff
@@ -110,4 +110,188 @@ inline static int32_t krml_time() {
...
+
+#define KRML_ACTUAL_FOR(i, z, n, k, x) \
+  do { \
+    for (uint32_t i = z; i < n; i += k) { \
+      x \
+    } \
+  while (0);
+
+#ifndef KRML_UNROLL_MAX
+#define KRML_UNROLL_MAX 16
```
franziskuskiefer, line 192: So 16 is the maximum. Is there a reason for that?

msprotz: It's just the default, I've updated the documentation of the option to explain the relationship between `-funroll-loops N` and `-DKRML_UNROLL_MAX=N'`

msprotz:
The reason I picked KRML_UNROLL_MAX to be 16 is that this guarantees that by default, if you do `-funroll-loops N`, then *all* the loops get unrolled by default. So this is a no-surprises policy.

PR #286, opened 2022-09-19
Stabilize auto-generated files such as ctypes.depend and .def files

msprotz:
As the name of the PR says. Note that this is not an alphabetical sort; rather, files are sorted according to their *C* filename (i.e., after bundling), and then declarations within each file are sorted alphabetically. But in any case, this should stabilize the output and avoid un-necessary diffs when refreshing HACL.


msprotz:
See https://github.com/hacl-star/hacl-star/pull/611 for motivation

krmllib/dist/generic/testlib.d.27147
```diff
@@ -0,0 +1,12 @@
+testlib.o: testlib.c TestLib.h \
```
R1kM, line 1: Should this file be included in the PR?

msprotz:
Thanks Aymeric!

PR #289, opened 2022-09-21
Test extraction of variant of Steel for_loop with a selector invariant

R1kM:
This PR adds support for a new Steel loop combinator, called `for_loop_full`, added in PR https://github.com/FStarLang/FStar/pull/2700.
This combinator is very similar to the existing Steel.Loops.for_loop, and so is the extraction.

This PR also adds an extraction test for `for_loop_full` in the existing SteelLoops file.


msprotz:
This looks good to me but is currently failing, probably because the corresponding F* PR has not yet been merged. 

```
 #14 99.50 [2022-09-21 13:23:47] [STDERR]SteelLoops.fst(19,2-19,15): (Error 72) Identifier not found: [for_loop_full]
#14 99.50 [2022-09-21 13:23:47] [STDERR]1 error was reported (see above)
#14 99.50 [2022-09-21 13:23:47] [STDERR]make: *** [Makefile:104: .cache/SteelLoops.fst.checked] Error 1
#14 99.50 [2022-09-21 13:23:47] [STDERR]make: *** Waiting for unfinished jobs....
```

Once the F* PR is merged, please:
- re-run actions here
- merge if you get a green.
Thanks!

msprotz:
One thought: is there any way we can have a type-polymorphic loop using a type class for the type of indices? That way we can have loops over uint32, uint64 and in the future size_t.

Might be nice to fix this design issue in Low*.

tahina-pro:
I just merged my alternative implementation into F*. Thus I reverted the Karamel extraction change, so this PR is only about _testing_ extraction of `for_loop_full`

PR #290, opened 2022-09-22
Support for compiling buffer casts (e.g. to_const) in WASM

msprotz:
```
Test #1 passed !
jonathan@absinthe:~/Code/everest/hacl-star/dist/wasm (polubelova_precomp_base) $ echo $?
0
```


polubelova (approved):
Thanks a lot!

PR #294, opened 2022-10-07
Add test for Steel.Array.malloca_of_list

R1kM:
Adds a small test for the `malloca_of_list` function proposed in https://github.com/FStarLang/FStar/pull/2696

This example extracts to the following C code
```C
int32_t test()
{
  uint8_t p[2U] = { (uint8_t)0U, (uint8_t)1U };
  uint8_t *res = p;
  uint8_t *res0 = res;
  uint8_t *r = res0;
  uint8_t *pt = r;
  uint8_t x = pt[0U];
  if (x == (uint8_t)0U)
  {
    uint8_t *p = r;
    KRML_HOST_FREE(p);
    return (int32_t)1;
  }
  else
  {
    uint8_t *p = r;
    KRML_HOST_FREE(p);
    return (int32_t)0;
  }
}
```


msprotz:
Looks good, but please fix build failures before merging, thanks

R1kM:
Currently fails with `free() : invalid pointer`. This is actually expected since `malloca_of_list` allocates on the stack, so the corresponding array shouldn't be freeable. This will need to be fixed directly in the F* model.

PR #295, opened 2022-10-07
Adding gcc4.4 exception

Frosne:
Summary: This PR request changes the behaviour of the target.h file for gcc4.4. 

GCC4.4 version complains about `KRML_CHECK_SIZE` and returns an exception ` #pragma GCC diagnostic not allowed inside functions. `
@protz proposed a working solution to have 
```
#  define _KRML_CHECK_SIZE_PRAGMA                                              \
    _Pragma("GCC diagnostic ignored \"-Wtype-limits\"")
```
for this compiler. The PR implements this solution.

### DIFF
--- a/include/krml/internal/target.h
+++ b/include/krml/internal/target.h
@@ -108,7 +108,7 @@ inline static int32_t krml_time() {
  * *elements*. Do an ugly, run-time check (some of which KaRaMeL can eliminate).
  */
 
-#if defined(__GNUC__) && __GNUC__ >= 4 && __GNUC_MINOR__ > 4
+#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ > 4))
 #  define _KRML_CHECK_SIZE_PRAGMA                                              \
     _Pragma("GCC diagnostic ignored \"-Wtype-limits\"")
 #else
--- a/include/krml/internal/target.h
+++ b/include/krml/internal/target.h
@@ -133,7 +133,7 @@ inline static int32_t krml_time() {
 #  define KRML_HOST_SNPRINTF(buf, sz, fmt, arg) snprintf(buf, sz, fmt, arg)
 #endif
 
-#if defined(__GNUC__) && __GNUC__ >= 4 && __GNUC_MINOR__ > 4
+#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ > 4))
 #  define KRML_DEPRECATED(x) __attribute__((deprecated(x)))
 #elif  defined(__GNUC__)
 /* deprecated attribute is not defined in GCC < 4.5. */
--- a/include/krml/internal/target.h
+++ b/include/krml/internal/target.h
@@ -108,7 +108,7 @@ inline static int32_t krml_time() {
  * *elements*. Do an ugly, run-time check (some of which KaRaMeL can eliminate).
  */
 
-#ifdef __GNUC__
+#if defined(__GNUC__) && __GNUC__ >= 4 && __GNUC_MINOR__ > 4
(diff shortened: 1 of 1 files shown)


msprotz (approved):
Thanks! To be clear, this disables the pragma for gcc <= 4.4

msprotz:
Unfortunately this breaks the build. 

```
13250 [2022-10-07 18:13:09] [STDOUT]cc -I. -I /home/opam/karamel/include -I /home/opam/karamel/krmllib/dist/minimal -Wall -Wextra -Werror -std=c11 -Wno-unused-variable -Wno-unknown-
13251 [2022-10-07 18:13:09] [STDERR]In file included from Bignum.h:15,
13252 [2022-10-07 18:13:09] [STDERR]                 from Bignum.c:8:
13253 [2022-10-07 18:13:09] [STDERR]Bignum.c: In function 'Impl_Bignum_add_alloc':
13254 [2022-10-07 18:13:09] [STDERR]/home/opam/karamel/include/krml/internal/target.h:121:24: error: comparison is always false due to limited range of data type [-Werror=type-limit
13255 [2022-10-07 18:13:09] [STDERR]  121 |     if (((size_t)(sz)) > ((size_t)(SIZE_MAX / (size_elt)))) {                  \
13256 [2022-10-07 18:13:09] [STDERR]      |                        ^
13257 [2022-10-07 18:13:09] [STDERR]Bignum.c:121:3: note: in expansion of macro 'KRML_CHECK_SIZE'
13258 [2022-10-07 18:13:09] [STDERR]  121 |   KRML_CHECK_SIZE(sizeof (uint32_t), dst_l);
```

@tahina-pro I think you initially wrote this (a3413b277cc3b61652aee34f5f5918b8f026c69c), do you know how to make it work both on the container's default compiler (which might be clang), and gcc, while avoiding tripping older GCCs?

tahina-pro:
@Frosne Should the condition rather be:
```c
#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ > 4))
```

tahina-pro:
@msprotz I believe the CI container's compiler is gcc (and the build may have failed because of the condition, as I suggested in my previous comment), but I don't know how other GCC-compatible compilers would interpret version numbers.
Regarding gcc < 4.4, I'm not sure we have much of a solution, per https://gcc.gnu.org/onlinedocs/gcc-4.4.4/gcc/Function-Specific-Option-Pragmas.html#Function-Specific-Option-Pragmas :
> The `#pragma GCC push_options` and `#pragma GCC pop_options` pragmas are not implemented in GCC versions earlier than 4.4.

msprotz:
The condition is copy-pasted from further down in the file (69ad32ef4c72bc75b9f4e06056f2570aa3325d6b, from @franziskuskiefer). @Frosne can you try Tahina's suggestion in both places? Thanks!

Frosne:
> @Frosne Should the condition rather be:
> 
> ```c
> #if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ > 4))
> ```

Yes, it also works. Updated the PR.

msprotz:
Ok, this is green. Great. Per my earlier comment:

> @Frosne can you try Tahina's suggestion in both places? Thanks!

Do you mind fixing the same test a few lines below? Thanks, and then we can merge.

Frosne:
Sorry, I missed that. Gonna do.

Frosne:
Hi,
I have missed a bit where is the second place you want the modifications to occur. It is here: https://github.com/FStarLang/karamel/blob/7dea9625a0f60faffca3c45fbf9a0a004b1e61c5/include/krml/internal/target.h#L136?

msprotz:
This is correct


~ jonathan

________________________________

msprotz (approved):
Looks great! Thanks

PR #296, opened 2022-10-10
Adds an example for model of top-level Steel arrays

R1kM:
A small example for the newly added Steel.TLArray


msprotz:
Looks good, please merge once you get a green

PR #297, opened 2022-10-13
Print toplevel comments with Doxygen syntax

msprotz:
In Microsoft mode, comments are printed with a special syntax so that they can be interpreted by documentation tools. This is actually a good idea, and we should do the same by default, on the basis that if a comment is not proper Doxygen syntax, nothing bad happens.

CCing @duesee and @franziskuskiefer who have requested this

### DIFF
--- a/src/PrintC.ml
+++ b/src/PrintC.ml
@@ -405,7 +405,7 @@ and p_stmt (s: stmt) =
 and p_stmts stmts = separate_map hardline p_stmt stmts
 
 let p_comments cs =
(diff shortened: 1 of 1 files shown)


franziskuskiefer (approved):
👍🏻 thanks

duesee:
A well-invested asterisk 🥂 Thanks!

msprotz:
CI seems to be dead. I'm going to "yolo this" and cross my fingers I don't have to revert.

PR #303, opened 2022-10-31
Extend support for size_t for arrays

R1kM:
This PR is the companion PR of https://github.com/FStarLang/FStar/pull/2741, which modifies the Steel array and loops libraries to use size_t as indices instead of uint32.

To support this in krml, this PR extends the krml checker to allow array indices to be either a uint32 or a size_t, and propagates the uint32 -> sizet changes to the Steel loops builtins.

It also adds SizeTArrays.fst, a small example of Steel arrays using size_t.


msprotz (approved):
This is exactly as we discussed, thanks! 👍

PR #304, opened 2022-11-09
Add C._zero_for_deref to list of builtins

R1kM:
This PR builds on top of #278, which added specific, more idiomatic extraction for reads on arrays.
The previous PR added a special value, C._zero_for_deref, which serves as a marker for more idiomatic extraction.
This PR marks this value as a builtin, to avoid adding an undefined reference to C._zero_for_deref when C is not in scope.

cc @tahina-pro

### DIFF
--- a/src/Builtin.ml
+++ b/src/Builtin.ml
@@ -410,7 +410,12 @@ let prepare files =
         name, snd f @ extra
       with Not_found ->
         f
-  ) files @ if List.mem_assoc "C" files then [] else [c_deref]
+  ) files @
+  (* This is unfortunately needed because of PR #278, and especially the corresponding
+     F* PR: References to module C can now occur even when the module is not in the scope.
+     If so, we add the definition that is needed as a builtin, since it will be rewritten
+     during C code generation *)
+  if List.mem_assoc "C" files then [] else [c_deref]
 
 let make_libraries files =
   List.map (fun f ->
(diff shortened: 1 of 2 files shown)


msprotz:
Aymeric and I chatted about this, and there is a fix (add C if not already in there, leave it and don't add the extra definition otherwise).

But overall, it's a pretty big violation of invariants for F* to generate a reference to the `C` module when it's not already in the dependency graph. Isn't there a cleaner solution? Some options Aymeric and I chatted about:
- when extracting Steel.X.read, generate a reference to Steel.X._zero_for_deref so that there's no extra module added in the graph
- add a wrapper in Steel.X, such as `inline_for_extraction let read = read' C._zero_for_deref` so that it appears directly in F*

Let us know?

tahina-pro:
> * when extracting Steel.X.read, generate a reference to Steel.X._zero_for_deref so that there's no extra module added in the graph

I think this is the cleanest option from the point of view of the user, but it is less tractable from the point of view of F* ulib developers, since they will need to add `Steel.X._zero_for_deref` into krml `src/CStarToC11.ml`, and potentially for every reference library.

> * add a wrapper in Steel.X, such as `inline_for_extraction let read = read' C._zero_for_deref` so that it appears directly in F*

I don't think this is possible: krmllib module C cannot be referenced from the F* ulib sources.

PR #306, opened 2022-11-16
Simplify Steel extraction

R1kM:
Following the fix to extraction in FStarLang/FStar#2754, this PR simplifies the build of Steel-related files.
It also tweaks the handling of Steel.Reference. The previous version was only correct because Steel_Reference.krml was always ignored, but led to duplicated libraries when adding it to Builtins.

### DIFF
--- a/src/Builtin.ml
+++ b/src/Builtin.ml
@@ -322,6 +322,7 @@ let hand_written = [
   lowstar_endianness;
   monotonic_hh;
   monotonic_hs;
+  steel_reference;
   steel_sizet_intros;
   hs;
   dyn;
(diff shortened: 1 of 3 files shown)


src/Builtin.ml
```diff
@@ -411,7 +411,7 @@ let prepare files =
         name, snd f @ extra
       with Not_found ->
         f
-  ) files @
```
msprotz, line 414: why don't you add steel_reference in hand_written, then? the behavior of the modules in hand_written is that they override anything otherwise extracted by F* (see line 405)

msprotz (changes requested):

msprotz (approved):

PR #307, opened 2022-11-16
Add support for -version flag

R1kM:
This PR extends krml with a new option, called -version, which displays the current version of KaRaMeL (as its current commit hash).
When krml is invoked with the -version, krml exits with return code 0 immediately after printing the version number.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ src/AutoConfig.ml:
 
 .PHONY: src/Version.ml
 src/Version.ml:
-	git rev-parse HEAD > $@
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -41,6 +41,10 @@ src/AutoConfig.ml:
 	  echo "let misc_dir = \"\";;" >> $@; \
 	fi
 
+.PHONY: src/Version.ml
+src/Version.ml:
+	@echo "let version = \"$(shell (git rev-parse HEAD))\"" > $@ \
```
msprotz, line 46: you don't need parentheses around git rev-parse HEAD

msprotz (approved):

PR #316, opened 2022-12-02
Make static-header and internal headers compose properly

msprotz:
Given a definition, there are two orthogonal concepts in krml.

- Visibility, a trivial lattice where private <= internal <= public.
  - private: definition is visible only to subsequent definitions in the current C file
  - internal: definition is not intended to be visible to end-users, but is visible to other files from this library; the definition is exported in internal/X.h
  - public: definition is visible to clients of this library; the definition is exported in X.h
- "static inline"-ness, controlled via the poorly-named `-static-header` option of krml
  - non-"static inline": definition goes in *the* C file; at C-compilation time, clients of the definition reference the symbol across translation units, to be resolved by the linker
  - "static inline": definition goes in *a* header, marked as "static inline"; at C-compilation time, every translation unit that includes the header gets their own copy of the definition per the C semantics

So, the intent of the `-static-header` option is to ensure that every client of a given definition gets a copy of the definition **in their translation unit** (almost always for the purposes of optimization).

Let's review some easy combinations of options:
- non-"static inline", private: function definition in the C file, marked as `static`
- non-"static inline", internal: function definition in the C file, function prototype in the internal .h file
- non-"static inline", public: function definition in the C file, function prototype in the public .h file

Previously, marking a definition as "static inline" (by passing a pattern to the poorly-named `-static-header` option that matches the definition) would automatically raise its visibility to public.

With this breaking change, we de-couple "static inline" from the visibility lattice, yielding the (new) following three combinations:
- "static inline", private: the user claims every translation unit that references the definition must get their own copy, but there is only one such translation unit, so no need to do anything special
- "static inline", internal: the definition is dropped from the C file; the complete definition for the function is exported in the internal header, with a "static inline" qualifier
- "static inline", public: the definition is dropped from the C file; the complete definition for the function is exported in the public header, with a "static inline" qualifier

In short, you can now have `static inline` definitions in internal headers, which is great for HACL* because we oftentimes need big constant arrays to be `static inline` for optimization (they are used by multiple files), but don't want to reveal their existence to clients (it's a complete implementation detail).

Clients that were relying on the previous behavior must update their bundle scripts, and replace `-static-header Foo` with `-static-header Foo -bundle Foo=` (force visibility of Foo to public) if `Foo` would be otherwise made private by a subsequent bundle (e.g. `Bar=Foo`)

I am particularly interested in feedback from @tahina-pro since I know that you are relying on this option for EverParse.


msprotz:
Just to be clear, this is a work in progress (several things are broken now). Just wanted to give a heads-up in case the plan outlined above breaks anyone's use-case.

msprotz:
Effect on HACL*: https://github.com/hacl-star/hacl-star/commit/3eed079dec17ea6227e64af59bed4f65aa0aadca

msprotz:
I forgot to mention that per @franziskuskiefer's request, this PR also changes the behavior of the cross-translation unit analysis: if a C function is marked inline (*not* static!) *and* is subjected to a call from another .c file, the inline gets automatically dropped with a warning.

(Prior to this PR, this behavior was only triggered when using `-fcompcert`)

PR #319, opened 2022-12-14
Small fix to debug info

msprotz:
This is useful for me as I need the topological order of headers for a script that tries to minimize header inclusion.


msprotz:
(Not intended to be on top of the other PR, but one more reason to merge it!)

PR #320, opened 2022-12-14
Do not include callconv.h by default anymore

msprotz:
This was used chiefly for miTLS. There's one detail about inline and antediluvian gcc versions that moved to target.h

The header will linger around (I have a patch to exclude it from dist/karamel in HACL* forthcoming) until it is removed for good.


msprotz:
This is one of the many small improvements that are on the radar for the Python PR

msprotz:
Got a local everest green, merging

PR #318, opened 2022-12-09
New internal node for null buffers.

msprotz:
A node `BufNull t` has type `t*`.

Backwards-compat desugaring of old application nodes in InputAst.ml

PR #324, opened 2023-01-05
New input AST node: EIsNull

tahina-pro:
#318 introduced a new node for null buffers, `EBufNull`. However, nothing was provided in the input AST (as opposed to the AST) to compare a pointer to NULL: polymorphic comparisons seem to be impossible in the input AST.
Thus, this PR introduces a new node for comparing a pointer to null, `EIsNull`.

### DIFF
--- a/src/InputAst.ml
+++ b/src/InputAst.ml
@@ -104,6 +104,7 @@ and expr =
   | EStandaloneComment of string
   | EAddrOf of expr
   | EBufNull of typ
+  | EIsNull of (typ * expr)
 
 and branches =
   branch list
--- a/src/InputAstToAst.ml
+++ b/src/InputAstToAst.ml
@@ -148,6 +148,7 @@ and mk_expr = function
   | I.EBufNull t ->
       { node = EBufNull; typ = TBuf (mk_typ t, false) }
 
+  | I.EIsNull (t, e)
   | I.EApp (I.ETApp (I.EQualified ( [ "LowStar"; "Monotonic"; "Buffer" ], "is_null"), [ t; _; _ ]), [ e ])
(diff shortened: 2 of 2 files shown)


msprotz:
Responding here for future reference rather than on the private Slack thread.

You do not need a new EIsNull node: it suffices to generate `EApp (ETApp (EOp (Eq, <whatever>), TBuf t); [ <x>; EBufNull t]))` to emit `<x> != null`. It is a quirk of the krml format that a type application of the equality operator is taken to mean polymorphic comparison, where the width argument to `EOp` is simply ignored (the `<whatever>` in my snippet above).

If you are adamant that you wish to add one extra node to the krml format, then I will request that you make sure `FStar.Extraction.Krml.fs` is consistent and uses EIsNull in all cases, so that the temporary workarounds in InputAstToAst.ml can eventually go away. Thanks!

msprotz (changes requested):
Per my comment above, either do without EIsNull, or make sure FStar.Extraction.Krml.fs uses EIsNull in all cases, to avoid confusion and having two ways of generating the same thing. Thanks.

tahina-pro:
Thanks a lot Jonathan!

> You do not need a new EIsNull node: it suffices to generate `EApp (ETApp (EOp (Eq, <whatever>), TBuf t); [ <x>; EBufNull t]))` to emit `<x> != null`. It is a quirk of the krml format that a type application of the equality operator is taken to mean polymorphic comparison, where the width argument to `EOp` is simply ignored (the `<whatever>` in my snippet above).

This is exactly the confirmation I needed. So let's shelve this PR for the time being. Thank you again!

PR #325, opened 2023-01-12
Primitive extraction of within_bounds

R1kM:
This PR is the companion PR of https://github.com/FStarLang/FStar/pull/2799, which adds a new primitive for restricted pointer comparison.

Most of the documentation is in the corresponding F* module, from karamel's perspective, this PR:
* Adds Steel.ArrayArith.within_bounds_ptr as a primitive builtin, with the signature (post-erasure) that it has in Steel.
* Adds a handwritten implementation of within_bounds_ptr, that is added to builtins.h. This implementation first casts the pointers to uintptr_t, as operations on pointers from different allocation units are undefined according to the C standard (section 6.5.8). Note, I added this function to builtins.h instead of steel_types.h to avoid bringing pthread in the context when concurrency is not required, but there might be a better place for this primitive.
* Adds a test demonstrating how to use and extract this primitive. Passing the options -static_header Steel.ArrayArith -no-prefix Steel.ArrayArith is required for linking to succeed.

### DIFF
--- a/include/krml/internal/builtin.h
+++ b/include/krml/internal/builtin.h
@@ -13,7 +13,7 @@
  * generate and try to link last a function with this type: */
 void krmlinit_globals(void);
 
-static inline within_bounds_intro(void *left, void *p, void *right) {
+static inline bool within_bounds_intro(void *left, void *p, void *right) {
   return (uintptr_t) left <= (uintptr_t) p && (uintptr_t) p <= (uintptr_t) right;
 }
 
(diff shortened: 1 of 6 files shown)


msprotz:
We have discussed this extensively and this is exactly as designed, looks good to me!

msprotz (approved):

R1kM:
After discussion with @msprotz , we decided to move the handwritten implementation of within_bounds_ptr to a separate header to avoid polluting `builtin.h` and existing clients with an unneeded definition. This header is however included in krmllib.h, which will allow its use by any client not using the -minimal option.

PR #326, opened 2023-01-12
Adds support for pointer substraction

R1kM:
This PR is the companion PR of https://github.com/FStarLang/FStar/pull/2800, which adds pointer substraction to Steel arrays.
To do so, this PR adds a new node, called EBufDiff to the karamel ASTs.
This node is extracted in C to a pointer substraction, and returns a ptrdiff_t.
There is a small example of its use in test/SizeTArray.fst

Note, extraction of pointer subtraction is currently not supported, and raises a failure. As ptrdiff is supposed to return the difference in number of elements, we would need a division operation in Wasm, which currently does not seem to be exposed/supported.

### DIFF
--- a/src/AstToCFlat.ml
+++ b/src/AstToCFlat.ml
@@ -802,7 +802,7 @@ and mk_expr (env: env) (locals: locals) (e: expr): locals * CF.expr =
       locals, CF.BufSub (e1, mk_mul32 e2 (mk_uint32 mult), base_size)
 
   | EBufDiff (e1, e2) ->
-      let mult, base_size = cell_size env (assert_buf e.typ) in
+      let mult, base_size = cell_size env (assert_buf e1.typ) in
       let locals, e1 = mk_expr env locals e1 in
       let locals, e2 = mk_expr env locals e2 in
       locals, mk_div32 (mk_sub32 e1 e2) (mk_mul32 (mk_uint32 mult) (mk_uint32 (bytes_in base_size)))
--- a/src/AstToCFlat.ml
+++ b/src/AstToCFlat.ml
@@ -467,6 +467,9 @@ let mk_minus1 e1 =
 let mk_plus1 e1 =
   CF.CallOp ((K.UInt32, K.Add), [ e1; mk_uint32 1 ])
 
+let mk_ptrdiff e1 e2 =
+  CF.CallOp ((K.PtrdiffT, K.Sub), [ e1; e2 ])
+
 let mk_memcpy env locals dst src n =
   let b = Helpers.fresh_binder ~mut:true "i" (TInt K.UInt32) in
   let locals, v, _ = extend env b locals in
@@ -794,6 +797,11 @@ and mk_expr (env: env) (locals: locals) (e: expr): locals * CF.expr =
       let locals, e1 = mk_expr env locals e1 in
       let locals, e2 = mk_expr env locals e2 in
       locals, CF.BufSub (e1, mk_mul32 e2 (mk_uint32 mult), base_size)
+
+  | EBufDiff (e1, e2) ->
+      let locals, e1 = mk_expr env locals e1 in
+      let locals, e2 = mk_expr env locals e2 in
+      locals, mk_ptrdiff e1 e2 
 
   | EBufWrite ({ node = EBound v1; _ }, e2, e3) ->
       let v1 = CF.Var (find env v1) in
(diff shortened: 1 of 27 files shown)


src/AstToCFlat.ml
```diff
@@ -795,6 +795,12 @@ and mk_expr (env: env) (locals: locals) (e: expr): locals * CF.expr =
       let locals, e2 = mk_expr env locals e2 in
       locals, CF.BufSub (e1, mk_mul32 e2 (mk_uint32 mult), base_size)
 
+  | EBufDiff (e1, e2) ->
+      let mult, base_size = cell_size env (assert_buf e.typ) in
+      let locals, e1 = mk_expr env locals e1 in
+      let locals, e2 = mk_expr env locals e2 in
+      locals, CF.BufDiff (e1, mk_mul32 e2 (mk_uint32 mult), base_size)
+
```
msprotz, line 803: why not just:
```
locals, mk_div32 (mk_sub32 e1 e2)
```
?

src/AstToCStar.ml
```diff
@@ -242,6 +242,8 @@ let rec mk_expr env in_stmt e =
       CStar.BufRead (mk_expr env e1, mk_expr env e2)
   | EBufSub (e1, e2) ->
       CStar.BufSub (mk_expr env e1, mk_expr env e2)
+  | EBufDiff (e1, e2) ->
+      CStar.BufDiff (mk_expr env e1, mk_expr env e2)
```
msprotz, line 246: same here, CStar is untyped, so why not go directly to a subtraction here?

msprotz: with the understanding that there is no need for division since the C compiler will do automatically

src/Checker.ml
```diff
@@ -359,6 +359,10 @@ and check' env t e =
       check_array_index env e2;
       check_buffer env t e1
 
+  | EBufDiff (e1, e2) ->
+      let t1, _ = infer_buffer env e1 in
+      check env (TBuf (t1, false)) e2
```
msprotz, line 364: use helper check_buffer

src/Checker.ml
```diff
@@ -617,6 +621,11 @@ and infer' env e =
       let t1, c = infer_buffer env e1 in
       TBuf (t1, c)
 
+  | EBufDiff (e1, e2) ->
+      let t1, _ = infer_buffer env e1 in
+      check env (TBuf (t1, false)) e2;
+      TInt PtrdiffT
```
msprotz, line 627: same here, `let t = infer env e1 in check_buffer t e2`

msprotz (changes requested):
Looks good but I think you can save yourself some work by emitting subtractions earlier

msprotz:
Another thought: might be worth having a WASM test. Something super simple:

```
int main() {
  uint64_t arr[8];
  return (arr + 4) - arr - 4;
}
```

just to make sure it does compute a proper number of elements. Add the file to WASM_FILES in the test directory and you should be good

src/AstToCFlat.ml
```diff
@@ -796,10 +802,10 @@ and mk_expr (env: env) (locals: locals) (e: expr): locals * CF.expr =
       locals, CF.BufSub (e1, mk_mul32 e2 (mk_uint32 mult), base_size)
 
   | EBufDiff (e1, e2) ->
-      let mult, base_size = cell_size env (assert_buf e.typ) in
+      let mult, _ = cell_size env (assert_buf e.typ) in
       let locals, e1 = mk_expr env locals e1 in
       let locals, e2 = mk_expr env locals e2 in
-      locals, CF.BufDiff (e1, mk_mul32 e2 (mk_uint32 mult), base_size)
+      locals, mk_div32 (mk_sub32 e1 e2) (mk_uint32 mult)
```
msprotz, line 808: is it `mult` or `base_size` you need? can't remember which is which

msprotz: I think you need to compute `(e1 - e2) / (mult * bytes_in base_size)`... if I remember well, `base_size` is the type or array cells (bytes, A64s, etc.) and `mult` is how many of these base cells a single element occupies... since `e1-e2` gives you a size in bytes, you need the more involved computation

msprotz (approved):
Looks good to me modulo fixing the wasm bit and writing a test to make sure it works

R1kM:
@msprotz I added the two wasm examples you were suggesting. Doing that, I realized that the extraction of size_t/ptrdiff_t that was added in an earlier PR was not supported in karamel. This PR also adds it by:
* Adding support to the Checker in the wasm case, for considering uint32_t and size_t as equivalent
* Extracting ptrdiff_t and size_t literals to int32 in wasm.

I now have a local green with these changes

PR #329, opened 2023-01-20
Ditch useless type abbreviations and fix #268

### LINKED ISSUE #268: Mangled type names for pointer to tuple with only one non-erased component
## What works
In general, Karamel is pretty good at erasing non-informative types away from tuple types. Consider for instance the following Low* code:

```fstar
module TH
module B = LowStar.Buffer
module HST = FStar.HyperStack.ST
module U8 = FStar.UInt8

let test (t: Type) = (t & unit)
let btest (t: Type) = B.buffer (test t)
```

<details>
    <summary>Click to expand</summary>

```fstar
let f (x: test U8.t) : Tot U8.t =
  match x with
  | (v, _) -> v

let g (x: btest U8.t) : HST.Stack U8.t
  (fun h -> B.live h x /\ B.length x >= 1) (fun _ _ _ -> True)
= let r = B.index x 0ul in
  let nres = f r in
  nres

let k (x: test (B.buffer U8.t)) : HST.Stack U8.t
  (fun h -> B.live h (fst x) /\ B.length (fst x) >= 1) (fun _ _ _ -> True)
= match x with
  | (v, _) -> let res = B.index v 0ul in res
```

</details>

Then, Karamel erases the `unit` tuple components and extracts to C code with the original type name as expected (albeit producing an unused type definition in between):

<details>
    <summary>Click to expand</summary>

```C
uint8_t TH_f(uint8_t x)
{
  return x;
}
uint8_t TH_g(uint8_t *x)
{
  uint8_t r = x[0U];
  uint8_t nres = TH_f(r);
  return nres;
}
typedef uint8_t *___uint8_t____;
uint8_t TH_k(uint8_t *x)
{
  uint8_t *v = x;
  return v[0U];
}
```

</details>

## What does not work
But using tuple types with non-informative components **both** inside and outside of a buffer type at the same time will produce mangled type names, obstructing code readability. From the following Low* code:

```fstar
let www (x: test (btest U8.t)) : HST.Stack U8.t
  (fun h -> B.live h (fst x) /\ B.length (fst x) >= 1) (fun _ _ _ -> True)
= match x with
  | (v, _) -> g v
```
Karamel produces the following C code (I merged the .h and .c files here for readability) :

```C
typedef uint8_t K___uint8_t___;
uint8_t TH_www(K___uint8_t___ *x)
{
  K___uint8_t___ *v = x;
  return TH_g(v);
}
```
Similar mangling also happens in this last example when replacing the tuple type with `dtuple2` or a custom record type, and replacing `unit` with any non-informative type (`squash`, `Ghost.erased`, etc.). Also, adding `inline_for_extraction` on `test`, `btest`, `f` and/or `g` does not help.

So my question is: why is Karamel producing and using this `K___uint8_t___` mangled type name instead of `uint8_t` as in the working cases above?

msprotz:
If the user refers to e.g. type `uint32_t & unit` in their code, this leads to a monomorphization (Monomorphization.ml):

```
type K__tuple_uint32_t_unit = { fst: uint32_t; snd: unit }

let incr (x: K__tuple_uint32_t_unit) = ...
```

Then, the unit field gets eliminated (DataTypes.ml):

```
type K__tuple_uint32_t_unit = { fst: uint32_t }

let incr (x: K__tuple_uint32_t_unit) = ...
```

Then, the record gets eliminated (actually it's a single-constructor inductive with a single field, but I'm sticking to the record here to make the explanations easier):

```
type K__tuple_uint32_t_unit = uint32_t

let incr (x: K__tuple_uint32_t_unit) = ...
```

Notice how the signature of `incr` doesn't change: this series of transformations is exclusively focused with the data type. This means that we end up with an ugly auto-generated type name that appears all over the place, when we could just use the underlying type instead (uint32_t).

This PR introduces a new flag for type declarations, called `AutoGenerated`. This flag is added at monomorphization-time, then caught later on to perform a second, more focused type abbreviation expansion phase, which unconditionally eliminates autogenerated type *abbreviations*.

@tahina-pro


msprotz:
I believe this fixes #268 but if you could confirm @tahina-pro that'd be great; thanks

msprotz:
@R1kM confirms that this also improves the code quality issues he's had with the Steel libraries, so merging this

PR #330, opened 2023-01-28
Positivity escape hatches for test cases

nikswamy:
This F* PR fixes a bug in the positivity checker that was not handling mutually inductive types correctly.
https://github.com/FStarLang/FStar/pull/2811

So, a few karamel test cases need no-positivity pragmas.

It seems that one of the examples is meant to be a minimalistic version of WireGuard? It's likely then that the actual F* sources of that application also needs this pragma ... at least until the Low* buffer type gains a positivity attribute.

### DIFF
--- a/test/DataTypesMut.fst
+++ b/test/DataTypesMut.fst
@@ -2,12 +2,14 @@ module DataTypesMut
 
 module B = FStar.Buffer
 
+#push-options "--__no_positivity"
 noeq
 type t1 a = a * a
 and t2 a =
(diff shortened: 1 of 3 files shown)


nikswamy:
Fwiw, many types in karamel already use this attribute

msprotz:
Thanks. I remember there were talks a long while ago of proving that B.buffer is positive based on the underlying model of addresses as nats... is it possible to do so now? Alternatively, is there any way to decorate the type to assert that it is positive (without necessarily proving it)?

PR #333, opened 2023-02-14
Update Makefiles to new dune-based F*

tahina-pro:
This is a companion PR to FStarLang/FStar#2815

With F* being built with `dune`, the structure and location of the F* libraries have changed: fstarlib and fstar-compiler-lib both become fstar.lib, and are now located in `$FSTAR_HOME/lib/fstar/lib/fstar_lib.cmx*` instead of `$FSTAR_HOME/bin/*.cmx*`.

Thus, this PR fixes the Makefiles of Karamel, krmllib and tests to invoke the OCaml compilers and linkers with the right libraries and with an appropriate `OCAMLPATH`.
In this process, this PR also fixes the way `FSTAR_HOME` is used, allowing Karamel to be built and tested even if the user installs F* via opam and does not set `FSTAR_HOME`.

This PR makes everest green (cf. project-everest/everest@925043549b2573160e93a23fab3abd10f969e70a)

Once this PR is approved, I'll merge it after merging the corresponding F* PR. Thanks in advance!

NOTE: This PR does NOT change the way Karamel is built. To track the conversion of Karamel itself to dune, please see #212 (in progress)


msprotz (commented):
This looks good, thanks. But the build of krml still depends on ocamlbuild, for tests/ctypes ... do you think this could be fixed as part of this PR? That way, we could completely deprecated ocamlbuild for the Everest build, which would be a huge win and would simplify maintenance going forward.

tahina-pro:
> This looks good, thanks. But the build of krml still depends on ocamlbuild, for tests/ctypes ... do you think this could be fixed as part of this PR? That way, we could completely deprecated ocamlbuild for the Everest build, which would be a huge win and would simplify maintenance going forward.

I know, but I still need to figure out how to do these tests/ctypes, and I don't want that hard piece to block the merge of the F* PR (which has been around for 3 weeks already.) So I'd like to merge this PR first, separately from #212, with separate PRs also in HACL* and elsewhere. This way, if anything breaks, we can more easily track whether it comes from the new F* dune build or the new Karamel dune build.

Thanks again!

PR #336, opened 2023-02-17
Eliminate spurious dependencies stemming from built-in definitions

msprotz:
I finally tracked down some superfluous #includes as part of https://github.com/hacl-star/hacl-star/pull/780

The short reason was: if a built-in operator (e.g. LowStar_BufferOps_op_Bang_Star) gets emitted in file X, then a file Y that needs the operator will #include the header for file X, even though the dependency is spurious: we ditch LowStar_BufferOps_op_Bang_Star before code-gen, meaning that Y doesn't really need X.

This reuses a definition (is_primitive) and skips dependency edges accordingly.


msprotz:
Screenshot 2023-02-17 at 8 00 01 AM

PR #340, opened 2023-04-04
Fix endianness issue in AIX

ayappanec:
AIX (always a big-endian) doesn't have `__BYTE_ORDER__` defined in any of its header files. GCC for this reason defines `__BYTE_ORDER__` in its own header files for AIX. But with XLC compiler, it's not the case.
The issue is exposed while compiling the latest python --> https://github.com/python/cpython/pull/102206

### DIFF
--- a/include/krml/lowstar_endianness.h
+++ b/include/krml/lowstar_endianness.h
@@ -96,7 +96,8 @@
 #  define le64toh(x) (x)
 
 /* ... generic big-endian fallback code */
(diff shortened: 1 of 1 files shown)


msprotz:
Thanks! I'll refresh the copy in HACL* and will pick up the changes upon my next PR for Python.

msprotz:
Let me know if there's anything else I can do to help.

ayappanec:
> Let me know if there's anything else I can do to help.

Thanks for all the support.

PR #341, opened 2023-04-05
Fix the sematics of Steel's spinlock

msprotz:
The Steel SpinLock type violates the value semantics of Low*. Its
low-level implementation, using pthread, relies on the address of the
value; therefore, `lock` is a type that cannot be passed by value (copying it is undefined). In order to maintain the "identity" of a lock
(the address of the value of type Steel_SpinLock_lock), we pass it by
reference, therefore guaranteeing that the value lives only in a single
place in memory.

This behavior should propagate to struct types that *contain*
Steel.Spinlock, so that a spinlock is not copied inadvertently by
copying a type that contains it. This PR adds a new mode to the existing analysis that passes structures by reference (krml's `-fnostruct-passing` option, used mostly for WASM these days) that passes a type by reference while making sure it's never copied.

To make this viable, I added a few peephole analyses that try to avoid un-necessary intermediary values (which would result in a fatal error on the basis that a value of type spinlock is being copied around).


msprotz:
Screenshot 2023-04-05 at 9 30 06 PM

nikswamy:
@R1kM @aseemr 

Do we really need this handwritten implementation of spin locks in krml?

At least for Steel.ST.SpinLock, we have a verified implementation that relies on the context instantiating an extern for a CAS. This is what we're using in Zeta.

Aymeric, are you using this spin lock in krml anywhere?

R1kM:
Hi Nik,

I'd be curious about concrete uses of your implementation, in particular how it works on different platforms.

Using spinlocks in userland is generally discouraged, which is why we are relying on the standard mutex library. This is already something we use in Starmalloc.

nikswamy:
We are using Steel.ST.SpinLock for code that runs in an enclave without necessarily any support for underlying libraries, like pthreads. The code is extracted with an extern for a CAS operation, which is provided directly in C and linked with the extracted code.

I can see that one would want to use more efficient and well-behaved mutexes. Maybe we shouldn't call this Steel_SpinLock, since it is not actually a spin lock. What about an interface like Steel.Primitive.PthreadMutex.fsti without any implementation in Steel, instead backed by this manual C implementation in karamel?

R1kM:
I'd be fine with renaming this into a Primitive.Mutex, that'd indeed be more descriptive and closer to what we are doing.
We should however wait until the current repo changes between Steel, F*, and karamel are stabilized to avoid conflicting changes.
I'll take care of the renaming once this is done.

tahina-pro:
> We should however wait until the current repo changes between Steel, F*, and karamel are stabilized to avoid conflicting changes. I'll take care of the renaming once this is done.

Indeed, we are in the process of moving Steel away from the F* repository, into its own repository: https://github.com/FStarLang/steel . As part of this operation, the current handwritten spinlock implementation is slated to move from `krmllib/c/steel_spinlock.c` to the new repository at https://github.com/FStarLang/steel/blob/main/src/c/steel_spinlock.c . I will open a Karamel PR soon from the `taramana_no_steel` branch, which is doing that move.

PR #344, opened 2023-04-12
add emscripten to endianness

franziskuskiefer:
Adding `__EMSCRIPTEN__` to endianness.

See https://github.com/cryspen/hacl-packages/pull/394


msprotz (approved):
Thanks!

PR #351, opened 2023-05-17
CI fixes

mtzguido:
I think this should fix the problem observed here https://github.com/FStarLang/karamel/actions/runs/4998273124, the slack post failed due to a null field.

Also update some dependencies while at it.


msprotz:
thanks!

PR #354, opened 2023-06-10
Makefiles: allow minimal build and cleaning without F*

mtzguido:
Hi Jonathan, this patch would allow to save some CI time by building karamel (`make minimal`) in parallel with F* and build the library only later, saving some time (though not much). Do you think it makes sense?

I could do the same thing for `clean`, avoiding the generation of `.depend`.

Also a tiny improvement to prevent against creating empty `.depend`s.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,10 @@ else
     FSTAR_HOME=$(dir $(FSTAR_EXE))/..
     OCAMLPATH:=$(OCAMLPATH)$(OCAMLPATH_SEP)$(FSTAR_HOME)/lib
   else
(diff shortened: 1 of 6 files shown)


msprotz (approved):
Thanks for the changes... looks good to me. I was unaware of output_deps_to ... is that recent? Anyhow, looks good, thanks!

mtzguido:
> I was unaware of output_deps_to ... is that recent?

Yup! Maybe about 2 months old, added precisely for this reason that a failing --dep call will create an empty .depend and trick make.

Merging then, thanks!

PR #355, opened 2023-06-18
krmllib: Makefile: fix ignore of non-extracted modules

mtzguido:
The makefile was wrongfully creating these files via the touch command, which meant to just update the timestamp. Use -c to prevent creation of files with this rule.

Also, for the modules which we do not want to extract, use --extract to leave them out of the dependency analysis, so we do not need to ignore them with the -bundle flag (which would also now fail, given that they do not exist).

I added the FStar.Stubs namespace to the list, as I will introduce that in soon-to-come F* PR.

### DIFF
--- a/krmllib/Makefile
+++ b/krmllib/Makefile
@@ -63,10 +63,15 @@ SHELL:=$(shell which bash)
 clean-c:
 	rm -rf dist out extract-all */*.o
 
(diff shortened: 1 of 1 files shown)


msprotz (approved):

msprotz:
Thanks Guido! this is a great cleanup

PR #356, opened 2023-06-22
actions: a more concise message for scheduled CI

mtzguido:
Github actions will, AFAIK, not provide any reasonable way of obtaining the commit message for a scheduled build. So just change the Slack message logic to print a short message with a success/failure and duration. It will anyway have a link to the relevant commit.

---

(This should get rid of the `<unknown>` in the slack messages in karamel-build.)


msprotz (approved):

msprotz:
Thanks Guido!

PR #357, opened 2023-06-26
New option to record renamings in a .h file

msprotz:
This simplifies name handling in the situation where a library and a client are both verified projects, extracting separately, and linking together.

Consider Library.A.Base1.fst and Library.A.Base2.fst, extracted with:

-bundle Library.A.Base1+Library.A.Base2=[rename=Library_A,rename-prefix]

The client, in order to use the library, passes `-library Library.*`, but without any further options, must also replicate the bundle above in order to generate code that references Library_A_foobar instead of Library_A_Base1_foobar; note that the latter would be a linking error.

This is onerous: the library must export its bundle options in a format consumable by the client; furthermore, the client is bound by the bundle choices of the library and has no way of grouping, say, in Library.h all of the `extern` declarations pertaining to Library while also at the same time respecting the renamings enabled by rename-prefix.

This PR offers an alternate route: the library records in a krmlrenamings.h file all of the name changes that deviate from the default, e.g.

```
#define Library_A_Base1_foobar Library_A_foobar
```

This allows the client to replace a litany of bundle/rename options with a mere `-add-include '"library/clients/krmlrenamings.h"'`


msprotz:
CC @pnmadelaine, whose life this should greatly simplify

msprotz:
Spurious failure for the red. The other green is 👍

msprotz:
As a side-effect, this breaks HACL* because it was relying on the -drop option which has been deprecated for many years now. The cross-referenced issue fixes it.

pnmadelaine:
@msprotz thank you very much for this!

pnmadelaine:
I use this in https://github.com/hacl-star/hacl-star/pull/789, everything seems to work fine! thanks again!

msprotz:
The test will be whether this allows *client* projects of HACL* (merkle-tree, everquic, noise*, and so on) to link against HACL* successfully without running into undefined symbols. But I want to merge this to prevent it from bit-rotting, so I'll send followup fixes for this particular feature as we find them.

PR #358, opened 2023-06-30
Some extraction tests for mutually recursive structs

tahina-pro:
What is the criterion for KaRaMel to properly extract, and C to properly compile, mutually recursive structures?

This PR adds a few test cases that give some answer to this question: as @msprotz said, when defining mutually recursive F* records, say `x1`, `x2`,..., `xn`, in this order, if `i < j <= k`, then `xi` can appear in the types of the fields of `xj`, but `xk` can appear in the types of the fields of `xj` only under `ref` or some pointer type.

### DIFF
--- a/src/Monomorphization.ml
+++ b/src/Monomorphization.ml
@@ -72,6 +72,17 @@ let monomorphize_data_types map = object(self)
   val mutable current_file = ""
   (* Possibly populated with something relevant *)
   val mutable best_hint: node * lident = (dummy_lid, []), dummy_lid
+  (* For forward references, a map from lid to its pending monomorphizations
+     (type arguments) *)
+  val pending_monomorphizations = Hashtbl.create 41
+  val seen_declarations = Hashtbl.create 41
+
+  method sanity_check () =
+    Hashtbl.iter (fun lid args ->
+      KPrint.bprintf "Missing monomorphization: %a<%a>\n" plid lid ptyps args
+    ) pending_monomorphizations;
+    if Hashtbl.length pending_monomorphizations > 0 then
+      Warn.fatal_error "Internal error: missing monomorphizations"
 
   (* Record a new declaration. *)
   method private record (d: decl) =
@@ -96,7 +107,7 @@ let monomorphize_data_types map = object(self)
       method! visit_TTuple () args =
         match Hashtbl.find state (tuple_lid, args) with
         | exception Not_found ->
-            let args = List.map (self#visit_typ ()) args in
+            let args = List.map (self#visit_typ false) args in
             TTuple args
         | _, chosen_lid ->
             TQualified chosen_lid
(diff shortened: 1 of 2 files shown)


tahina-pro:
I just pushed a test that satisfies the condition we mentioned above:

https://github.com/FStarLang/karamel/blob/0df4ef66155bb0760b3093f40165d5e320040f88/test/MutualStruct.fst#L93-L113

but that fails to compile:
```
✘ [CC,.output/MutualStruct.out/MutualStruct.c]
In file included from .output/MutualStruct.out/MutualStruct.c:8:
.output/MutualStruct.out/MutualStruct.h:69:11: error: field ‘object7_pair_fst’ has incomplete type
   69 |   object7 object7_pair_fst;
      |           ^~~~~~~~~~~~~~~~
.output/MutualStruct.out/MutualStruct.h:70:11: error: field ‘object7_pair_snd’ has incomplete type
   70 |   object7 object7_pair_snd;
      |           ^~~~~~~~~~~~~~~~
```
because KaRaMeL generated the definition of `object7_pair` before `object7`, which is too soon.

msprotz:
@tahina-pro do you mean that your test should succeed but doesn't currently?

tahina-pro:
> do you mean that your test should succeed but doesn't currently?
Yes, I expected karamel to produce the forward-declarations in the same order as in the source file, for the produced C file to properly compile.

tahina-pro:
To solve this issue, given a sequence of mutually inductive type definitions, would it be simpler for Karamel to:
1. Check the criterion mentioned in the description of this PR.
2. Then, produce forward-declarations for all types in this sequence, in the same order.
3. Then, produce all actual type definitions, in the same order.
?

msprotz:
I actually looked into this. What happens is that krml reorders definitions as-needed, for the case where you wrote something that does not need forward references but needs reordering in order to compile. For instance:

```
type foo = | Foo: bar -> foo

and bar = | Bar: baz -> bar

and baz= ...
```

If I had to guess, I'm pretty sure that this was because of miTLS. I don't want to break backwards-compat, so what I'm thinking of doing is reordering only when *not* going behind a pointer. If going behind a pointer, a forward reference is emitted, then the relevant type is visited later on, hence preserving the order you need.

I need to try it out (I have a partial patch locally)

msprotz:
I have a tentative fix, which has the interesting side-effect that it generates *more* forward declarations. For instance, in test/Wireguard.fst, following the reordering dictated by the traversal *does* generate correct C declarations without the need for forward declarations.

I could optimize to not generate *too many* forward declarations. Right now krml doesn't keep track of which have been emitted already.

msprotz:
(everest regression running)

msprotz:
Screenshot 2023-07-05 at 10 27 14 PM

tahina-pro:
Thanks a lot Jonathan! Your fix works very well!
With that, Karamel consistently produces expected forward declarations for each type xk used in xj under a pointer type constructor. (Karamel seems to produce 2 of them, but that's fine for now.)

PR #361, opened 2023-07-24
Bring anonymous unions into the data type monomorphization / reordering phase

msprotz:
@tahina-pro can you confirm this works for you?

With SteelC now generating anonymous unions, krml needs to consider anonymous union type declarations in the phase that:
- emits type declarations, while
- monomorphizing them, and
- emitting forward declarations in case there is mutual recursion
- and preserving the order so long as the declarations are referred to behind pointers.

### DIFF
--- a/krmllib/hints/C.Endianness.fst.hints
+++ b/krmllib/hints/C.Endianness.fst.hints
@@ -8,7 +8,7 @@
       1,
       [ "@query" ],
       0,
-      "c169be3b83d65c25442276d4189d1094"
+      "c9d34bcf6778bfc0790376dfadcef8db"
     ],
     [
       "C.Endianness.load16_le",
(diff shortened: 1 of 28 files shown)


tahina-pro (approved):
Thanks a lot Jonathan! SteelC tests from FStarLang/steel@e71eda7168172a3391d1bc0a5d272599b50233e6 succeeded, so we can go ahead and merge. Thank you again!

(Ideally, I should move those tests to the Karamel test suite, since they already work on Steel main, but I'll do that only once SteelC is deemed stable enough. Right now only SteelC is exercising Karamel untagged unions, so SteelC is the only way to test such features.)

msprotz:
Great. And I had an Everest green, too, so we're all good. Glad this fixed your issue!

I will happily take a PR that moves the testcases back to krml once it makes sense to do so. Just let me know. Cheers.

PR #364, opened 2023-08-18
Add support for polymorphic external definitions

msprotz:
This PR allows krml to internally handle external definitions that are type-polymorphic, compiling them an application of a the types, followed by the value arguments.

The chief use-case is to implement things in C that fall outside of the Low* model. Consider, for instance, Lib.Memzero, a function that in Low* has the signature:

```
val memzero: #t -> b:buffer t -> l:uint32 { l == len b } -> unit ...
```

The Low* functions takes a number of elements, but the C safe memory scrubbing functions take a number of *bytes*. Until now, there was ad-hoc support in krml to handle this very specific function; now, with this PR, `memzero` is just a regular type-polymorphic external declaration, and calls to memzero become:

```
memzero(t, b, l);
```

e.g.

```
uint32_t x[4];
memzero(uint32_t, x, 4);
```

this then allows the user to define `memzero` by hand in a C header instead of adding hardcoded support in the krml source code. For instance:

```
#define memzero(t, x, l) explicit_bzero(x, sizeof t * l);
```


msprotz:
There are a variety of other changes (tightening of the internal checker, fixing a few shortcomings in some internal nodes like EAbort, etc.). Another notable change is the switch to a canonical "opam" layout with src and lib, so heads-up to @tahina-pro and @R1kM in case this impacts packaging.

Finally, I'll note that this is not enabled by default; right now, there is an option to allow type applications, or if the option is unused, a whitelist of things that are allowed to be external *and* type-polymorphic (right now, only memzero and the steel spinlock functions are in that list). One possible avenue to ease the transition would be to "recycle" the CMacro attribute, and interpret it, when applied to a definition without an implementation, to mean that this function is safe to compile as a type application. Or to have another attribute to indicate that this is ok to code-gen with a type application.

PR #366, opened 2023-08-20
Handle a new CNoInline attribute 

msprotz:


include/krml/internal/target.h
```diff
@@ -57,6 +57,18 @@
 #  define KRML_HOST_IGNORE(x) (void)(x)
 #endif
 
+#ifndef KRML_NOINLINE
+#  if defined(_MSC_VER)
```
karthikbhargavan, line 61: Is it true that LLVM also sets _MSC_VER on windows? Maybe we need some more flags conditions here?

include/krml/internal/target.h
```diff
@@ -57,6 +57,18 @@
...
+#    define KRML_NOINLINE __declspec(noinline)
+#  elif defined (__GNUC__)
```
karthikbhargavan, line 63: clang also sets `__GNUC__`, but we could use `|| __llvm__` to be sure to cover both.

include/krml/internal/target.h
```diff
@@ -57,6 +57,18 @@
...
+#    define KRML_NOINLINE __attribute__((noinline))
+#  else
+#    warning "The KRML_NOINLINE macro is not defined for this toolchain!"
```
karthikbhargavan, line 66: In this third case, we could simply define `KRML_NOINLINE` to be empty with the warning?

include/krml/internal/target.h
```diff
@@ -151,202 +167,202 @@ inline static int32_t krml_time(void) {
...
-#define KRML_LOOP1(i, n, x) { \
-  x \
-  i += n; \
-}
+#define KRML_LOOP1(i, n, x)                                                    \
+  { x i += n; }
 
-#define KRML_LOOP2(i, n, x) \
-  KRML_LOOP1(i, n, x) \
+#define KRML_LOOP2(i, n, x)                                                    \
```
karthikbhargavan, line 192: What's the reason for these spaces?

include/krml/internal/target.h
```diff
@@ -151,202 +167,202 @@ inline static int32_t krml_time(void) {
...
+#define KRML_ACTUAL_FOR(i, z, n, k, x)                                         \
+  do {                                                                         \
+    for (uint32_t i = z; i < n; i += k) {                                      \
+      x                                                                        \
+    }                                                                          \
   } while (0)
 
 #ifndef KRML_UNROLL_MAX
-#define KRML_UNROLL_MAX 16
+#  define KRML_UNROLL_MAX 16
```
karthikbhargavan, line 266: Why a space between # and define?

karthikbhargavan (changes requested):

include/krml/internal/target.h
msprotz: this is clang-format's default settings, I stupidly re-formatted this file after my changes without realizing it would generate a lot of diff noise... please ignore

include/krml/internal/target.h
msprotz: same, this is clang-format's default settings

include/krml/internal/target.h
msprotz: done

include/krml/internal/target.h
msprotz: I'm not sure, but all the macros in this file follow the convention that _MSC_VER means "MSCV-compatible" (and presumably, should clang define this macro, then clang would be able to parse __declspec rather than __attribute__), and __GNUC__ means gcc/clang...

If that's not right then we have a bigger problem than just this macro

karthikbhargavan (approved):

PR #373, opened 2023-09-05
Fix install target

msprotz:


msprotz:
reported by @pnmadelaine

PR #374, opened 2023-09-18
NOINLINE improvement

karthikbhargavan:
This PR simplifies the KRML_NOINLINE fix for masking functions to use the attribute "unused" instead of pragmas.


msprotz (approved):
This is great, thanks for finally figuring this out!

PR #375, opened 2023-09-20
Various improvements / fixes, some reported by Tahina

msprotz:


tahina-pro (approved):
Thanks a lot Jonathan!

With this PR, I can now have SteelC implement `Steel.ST.C.Types.void_ptr` as `void*` with `TAny` instead of resorting to an explicit `.h` file defining that type. Thanks to this PR, fields of `TAny` type are no longer erased, and they appear with the `void*` type.

More precisely, with this Karamel PR, I have a Steel branch where I define the following SteelC code:

https://github.com/tahina-pro/steel/blob/435c90c10d4c76e13235b44146a281943252277a/share/steel/examples/steelc/MiniAny.fst#L1-L12

```fstar
module MiniAny

noeq
type test = {
  a: FStar.UInt32.t;
  b: Steel.ST.C.Types.Base.void_ptr;
}

let x
  (f: test)
: Tot Steel.ST.C.Types.Base.void_ptr
= f.b
```

and I run `make MiniAny.krml && krml -warn-error @4-26 -skip-linking MiniAny.krml` to extract it into the following C code:
```c
typedef struct MiniAny_test_s
{
  uint32_t a;
  void *b;
}
MiniAny_test;
void *MiniAny_x(MiniAny_test f)
{
  return f.b;
}
```

PR #381, opened 2023-09-29
Ignore all unused variables automatically

msprotz:


tahina-pro:
Thanks a lot Jonathan!

PR #382, opened 2023-10-02
Implicit conversions in  fstar_uint128_gcc64.h

franziskuskiefer:
Should this be done in the msvc version as well?


msprotz:
I think so, yes. Thanks!!

franziskuskiefer:
> I think so, yes. Thanks!!

msvc should be fine https://github.com/FStarLang/karamel/blob/f8d97a5edd60e392e977c8e0aafbd530fe9caa5c/krmllib/c/fstar_uint128_msvc.h#L354

msprotz:
Cool, thanks!

PR #384, opened 2023-10-06
Insert KRML_IGNORE for variables that may end up being unused due to ifdefs; same deal with static (private) functions

msprotz:


msprotz:
Screenshot 2023-10-09 at 9 31 27 AM

PR #386, opened 2023-10-10
Update builtin.h to account for Solaris

msprotz:


msprotz:
according to https://sourceforge.net/p/predef/wiki/OperatingSystems/ `sun` is defined on Solaris

PR #387, opened 2023-10-11
Address long-standing issue with right shift -- fixes #102

### LINKED ISSUE #102: Missing cast in left operand of right shift
The following program outputs `z = 255` despite `assert (z == 1uy)` being provable in F*:
```F#
module M

open FStar.HyperStack.All
open FStar.UInt8

val discard: bool -> St unit
let discard _ = ()

let print s = discard (IO.debug_print_string s)

val main: unit -> St C.exit_code
let main _ =
  let z = (0uy -%^ 1uy) >>^ 7ul in
  assert (z == 1uy);
  print (Printf.sprintf "z = %uy\n" z);
  C.EXIT_SUCCESS
```

(compiled with `krml -no-prefix M -add-include '"extracted/FStar_String.h"' -drop WasmSupport M.fst -o M.exe`.)

The generated C code uses
```uint8_t z = (uint8_t)0U - (uint8_t)1U >> (uint32_t)7U;```
while I think it should use
``` uint8_t z = (uint8_t)((uint8_t)0U - (uint8_t)1U) >> (uint32_t)7U;```

Note also that the C11 standard says that the semantics of a right shift with a negative left operand is implementation-defined:

_The result of E1 >> E2 is E1 right-shifted E2 bit positions. If E1 has an unsigned type or if E1 has a signed type and a nonnegative value, the value of the result is the integral part  of  the  quotient  of E1/2^E2. If E1 has a signed type and a negative value, the resulting value is implementation-defined._

### PR DESCRIPTION
**Pull request by msprotz**
After several days and many discarded designs, I think I have found a reasonable way to i) optimize the generation of integer arithmetic constants and computations (hi @franziskuskiefer, @karthikbhargavan, please take a look at HACL branch protz_shift to see the effects of the change), ii) fix the semantics of right-shift (hi @s-zanella, @polubelova), and iii) in the process of doing so, fix integer conversion warnings for right-shifts (there are still spurious warnings in a lot of places in HACL, but at least that one reported on Slack by @franziskuskiefer in FrodoKEM is now gone).

I know that others have encountered #102 in one form or another in the past, so my request to @tahina-pro: can you please take a careful look at this PR and tell me whether you believe this compilation scheme is correct or not? In addition, can you see if this helps get rid of your earlier workarounds?

Thanks!

### DIFF
--- a/lib/AstToCStar.ml
+++ b/lib/AstToCStar.ml
@@ -323,8 +323,19 @@ and mk_arith env e =
       | _ ->
           assert false
       end, false
-  | _ ->
+  | EConstant _ ->
+      (* Constants are emitted with the U suffix which preserves the invariant
+         that every subexpression operates over unsigned int until the final
+         cast, or until a mask is needed to preserve semantics. *)
       mk_expr env false e, true
+  | _ ->
+      (* Something else. Who knows? Maybe a function call, a field reference, a
+         variable, a global... which will be upcast into `int`, which is *not*
+         what we want! (See compilation strategy above.). We cast. *)
+      if e.typ = TInt UInt16 || e.typ = TInt UInt8 then
+        CStar.Cast (mk_expr env false e, Int UInt32), true
+      else
+        mk_expr env false e, true
 
 and mk_expr env in_stmt e =
   let mk_expr env e = mk_expr env false e in
--- a/lib/AstToCStar.ml
+++ b/lib/AstToCStar.ml
@@ -219,6 +219,113 @@ let rec unit_to_void env e es extra =
   | _ ->
       CStar.Call (mk_expr env e, List.map (mk_expr env) es @ extra)
 
+(* This function focuses on the compilation of arithmetic expressions to take
+  into account implicit promotions to `int` (per the C standard) of types
+  smaller than int. We make the following assumption: int = 4 bytes, short = 2
+  bytes, char = 1 byte, and the unsigned versions of theses types are called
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by tahina-pro — APPROVED** [review]

File: lib/AstToCStar.ml [hunk 1]
```diff
@@ -219,6 +219,127 @@ let rec unit_to_void env e es extra =
...
       CStar.Call (mk_expr env e, List.map (mk_expr env) es @ extra)
 
+(* This function focuses on the compilation of arithmetic expressions to take
+  into account implicit promotions to `int` (per the C standard) of types
+  smaller than int. We make the following assumption: int = 4 bytes, short = 2
+  bytes, char = 1 byte, and the unsigned versions of theses types are called
+  uint32_t, uint16_t, and uint8_t. Yes, the C standard doesn't impose these
+  choices, but all systems that we care about obey these restrictions
+  (https://www.viva64.com/en/t/0012/) and no, we don't intend to run the
+  generated C code on a PDP-11.
```
tahina-pro, line 229: It would be nice to document those hypotheses in, say, the Low* tutorial.
Thanks Jonathan!

msprotz: done in 2640744

msprotz:
Note that in Wasm every subexpression of type smaller than uint32 is masked

https://github.com/FStarLang/karamel/blob/da83fc312e2f3833f5746dfb702a456048a7a3da/lib/CFlatToWasm.ml#L621-L634

so nothing to fix there

PR #391, opened 2023-10-22
Fix karamel.opam: add ctypes and ctypes-foreign

cmovcc:


msprotz:
Build fail, please try to see with @pnmadelaine what's up

msprotz:
@pnmadelaine doesn't know what's going on -- @tahina-pro do you have a clue?

msprotz:
Maybe @mtzguido knows?

mtzguido:
I think it's jump OPAM defaulting to `no` when installing dependencies. Trying..

mtzguido:
Sigh I think this is a discrepancy between the standalone build and the hierarchic one. The standalone build already does set OPAMYES=1, so the previous CI job from the original commit should have worked. The ones for my pushes just now are hierarchical (since there's an F* image on the machine), and seem to work even before my change (the merge is green).

I think the proper thing to do is removing the two flavors and just have one, but unsure on what that actually looks like if we want it to be fast. Also this should at least work around the problem if it shows up again.
```
diff --git a/.docker/standalone.Dockerfile b/.docker/standalone.Dockerfile
index d92931a0..b7c20531 100644
--- a/.docker/standalone.Dockerfile
+++ b/.docker/standalone.Dockerfile
@@ -12,6 +12,8 @@ RUN apt-get update && \
       git \
       gawk \
       coreutils \
+      libffi-dev \
+      pkg-config \
       sudo \
       opam \
       python3 \
```

msprotz:
@cmovcc can I let you add Guido's patch to this PR? thanks!

mtzguido:
I hit this today with the new CI: I had to install these two manually:
https://github.com/FStarLang/karamel/blob/92d7cfa1df526d4f5e814b27e0d16a1ed2e68494/.github/workflows/ci.yml#L58-L59

From what I can tell, ctypes and ctypes-foreign are not really needed for krml, but only for the tests, so I'm not sure if they should be made dependencies. It would probably be fine, since they are small packages, so no strong opinion.


The dockerfiles can be removed I think, see #641.

PR #397, opened 2023-12-04
Generalization of the "pass by ref" for the case where functions are returning arrays (not pointers)

msprotz:
Rust makes a proper distinction between arrays and pointers, and functions in Rust can return arrays (by value). This can of course be optimized into a caller allocating space for the array (known at compile-time) and the callee copying its return value into the destination address.

There was already a phase for that in krml (used for mitls, a long time ago, but also for the wasm backend) so it was mostly a matter of tweaking it.


msprotz:
(got a local Everest green)

PR #399, opened 2023-12-11
Get rid of the compatibility `Option` module

Niols:
This PR gets rid of the local `Option` module. The standard library contains an `Option` module since OCaml 4.08, which contains all what this compatibility module was used for. To be precise, the module in question provided:

- `must`: replaced by `Stdlib.Option.get`
- `map`: provided by `Stdlib.Option`
- `map_or`: unused
- `or_empty`: replaced by `Stdlib.Option.value ~default:""`
- `is_some`: unused and provided by `Stdlib.Option`


msprotz:
Awesome! Thank you

msprotz:
Not sure what is your particular interest here, but if you're into that sort of cleanups, there might be more functions in KList that now have equivalents in OCaml's standard library

PR #400, opened 2023-12-13
Cleanup the `KList` module

Niols:
This PR cleans up the `KList` module by using newly-introduced `Stdlib.List` utilities:

- `KList.filter_map` is `Stdlib.List.filter_map`
- `KList.map_flatten` is `Stdlib.List.concat_map`
- `KList.make` is `Stdlib.List.init` **except for the side-effects** (I think this is fine but you might want to check this.)
- `KList.find_opt` is `Stdlib.List.find_map`
- `KList.assoc_opt` is `fun k -> Stdlib.List.assoc (Some k)`

Additionally:

- We get rid of `KList.max` and `KList.min` which are easily written using the already-existing `KList.reduce`.
- Some functions will become available in OCaml 5.1 and are commented as such to help future cleanup.
- We rewrite `KList.fold_lefti` as a special case of `List.fold_left`, hopefully making it more readable.
- We rewrite `KList.last` as a recursive function to make it more performant.

The commits are independent and I will happily revert/remove any commit that you don't find useful.

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -625,7 +625,7 @@ end
 (** More helpers *)
 
 let filter_decls f files =
-  List.map (fun (file, decls) -> file, KList.filter_map f decls) files
+  List.map (fun (file, decls) -> file, List.filter_map f decls) files
 
 let map_decls f files =
   List.map (fun (file, decls) -> file, List.map f decls) files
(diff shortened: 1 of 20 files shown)


msprotz:
Looks excellent! Thanks so much. I don't think I have any side effects in KList.init so it's safe to change the evaluation order.

PR #401, opened 2023-12-14
remove backtracking in `reduce_full_match`

amosr:
this PR (hopefully) addresses a performance / non-termination problem I was having.

I had a problematic program contains many matches where the scrutinee itself contains nested matches (eg `match match match match match match t with ... with ... with ...`). in reduce_full_matches, it reduces the scrutinee and the branch, then if it can't apply case-of-constructor it backtracks, which ends up having to reduce the scrutinee and branch again separately. in my case with lots of nested matches, this seemed to be taking exponential time in the number of nested matches.

### DIFF
--- a/lib/DataTypes.ml
+++ b/lib/DataTypes.ml
@@ -1207,7 +1207,6 @@ let remove_full_matches = object (self)
   inherit [_] map as super
 
   method! visit_EMatch (_, t as env) c scrut branches =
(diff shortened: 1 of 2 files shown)


msprotz (approved):
Ok looks good. Thanks!

amosr:
ah, I think my test case was wrong (it was returning -1). hopefully fixed now.

PR #405, opened 2024-01-17
feat(nix): expose `krml` library in `passthrough`

W95Psp:
This PR ships the `krml` library in the Nix build, so that it can be consumed by other OCaml packages through Nix.


msprotz (approved):
Excellent ! Thanks so much Lucas.

PR #406, opened 2024-02-02
Nano-pass: remove empty records

msprotz:
As it turns out, the case where a struct is made of *only* ghost fields had previously never been exercised before. This is easily fixed with another nano-pass that performs a little bit of cleanup.


msprotz:
Reported by @R1kM

PR #408, opened 2024-02-13
Recognize more cases of zero-initialized arrays. Refine hoisting phas…

msprotz:
…e to be nested-arrays aware.


msprotz:
Nested arrays are not representable in Low*, but they are in Rust, so this makes a couple things aware of that new fact for the Eurydice use-case.

PR #416, opened 2024-03-16
Extend Rust extraction support for ConstBuffer in subexpression positions

R1kM:
This is a small tweak to the existing fallback for the case when a sub operation appears in subexpression position (e.g., as an argument to a function call), which happens in HACL* when using inline_for_extraction for sub operations.

The previous behavior was to always take a mutable borrow on the range created by the subbuffer. When using const buffers, this led to borrow-checking issues, where we tried to take a mutable borrow on a shared borrow. This PR slightly tweak this behavior to check (based on the argument type) whether the buffer argument is const. In this case, it is extracted as a shared borrow.

A small example (Rust6) showcases the problematic pattern, which is present in HACL*'s Ed25519/P256/K256 when using precomputed tables, where the function `f` was previously extracted to
```rust
pub fn f(b: &[u32]) -> ()
{
    let mut res: [u32; 2] = [0u32; 2usize];
    ((&mut res)[0usize..2usize]).copy_from_slice(&(&mut b[0usize..])[0usize..2usize])
}
```
This PR removes the inner mutable borrow of b.

### DIFF
--- a/lib/AstToMiniRust.ml
+++ b/lib/AstToMiniRust.ml
@@ -873,9 +873,13 @@ and translate_expr_with_type (env: env) (e: Ast.expr) (t_ret: MiniRust.typ): env
       (* This is a fallback for the analysis above. Happens if, for instance, the pointer arithmetic
          appears in subexpression position (like, function call), in which case there's a chance
          this might still work! *)
+      let is_const_tbuf = match e1.typ with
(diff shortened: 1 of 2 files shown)


msprotz (approved):

msprotz:
Wonderful! Nice detective work and perfect fix.

PR #421, opened 2024-04-10
Fix an interaction between polymorphic type applications (LowStar.Ign…

msprotz:
…ore) and passing structs by ref


pnmadelaine:
@msprotz the Nix build succeeds on FStar/master, it was also fine yesterday (on 695c472)

I don't know much about the setup to build F* from the Karamel CI, but it looks like the problem is located there!

pnmadelaine:
> Weird. It suddenly started working?!!

I learned better than to ask questions when something like this happens.

PR #426, opened 2024-04-23
Fix parentheses for MSVC -- triggers compiler warnings

msprotz:


msprotz:
CC @mlindgren

mlindgren:
Confirmed via local build that this fixes the warnings for me

PR #431, opened 2024-05-23
Support for monomorphization over regular values like trait methods

msprotz:
This is in support of https://github.com/AeneasVerif/eurydice/pull/11 which needs to build against this.

Both PRs will be merged at the same time; basically, this adds the option to monomorphize over an additional set of values that do not pertain to const generics and therefore are not bound in types (like const generics).


msprotz:
Got a local Everest green (after disabling miTLS extraction, per discussion)

PR #437, opened 2024-06-12
Revert "Substitutions are not type-preserving in C"

msprotz:
Reverts FStarLang/karamel#436


msprotz:
This caused regressions in MLkem, will redo later.

PR #438, opened 2024-06-14
Fix spinlock extraction due to stronger monomorphization

R1kM:
#431 introduced a stronger monomorphization, which in particular rewrites occurences of Steel_SpinLock_lock () into Steel_SpinLock_s_lock. This in turn leads to escaping the address-passing semantics of locks.

This PR fixes this by adding Steel_SpinLock_s_lock to the specific handling of locks.


msprotz:
nice find -- great!!

PR #441, opened 2024-06-27
Attempt to unify implicit and explicit lifetime declarations

R1kM:
As part of the Rust extraction of HACL*, several functions are not extracted with messages such as the following:
```
translating Hacl.Streaming.Blake2b_32.reset_raw: Krml.Warn.Fatal("type mismatch;\n  e=(@7).fst\n  t=&'a mut [hacl::hash_blake2b::blake2_params <'a>]\n  t_ret=&mut [hacl::hash_blake2b::blake2_params]\n  x=@7.fst")
```

The reason is that there is a mismatch between the inferred and the expected type, as the inferred type includes explicit lifetime annotations.
This PR aims to provide a best-effort workaround, by comparing types with lifetime annotations erased.
The soundness of this comparison will be left to the Rust borrow-checker, in particular, incorrect cases where there would truly be a mismatch (e.g., because of different lifetimes in `t` and `t_ret`, or because of multiple lifetimes) will be extracted by karamel with this patch, but will fail Rust compilation.

### DIFF
--- a/lib/AstToMiniRust.ml
+++ b/lib/AstToMiniRust.ml
@@ -589,6 +589,19 @@ and translate_array (env: env) is_toplevel (init: Ast.expr): env * MiniRust.expr
 (* However, generally, we will have a type provided by the context that may
    necessitate the insertion of conversions *)
 and translate_expr_with_type (env: env) (e: Ast.expr) (t_ret: MiniRust.typ): env * MiniRust.expr =
+  let rec erase_lifetime_info (t: MiniRust.typ) = match t with
+    | Ref (_, bk, t) -> MiniRust.Ref (None, bk, erase_lifetime_info t)
+    | Vec t -> Vec (erase_lifetime_info t)
+    | Array (t, n) -> Array (erase_lifetime_info t, n)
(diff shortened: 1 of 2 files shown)


lib/AstToMiniRust.ml
```diff
@@ -589,6 +589,19 @@ and translate_array (env: env) is_toplevel (init: Ast.expr): env * MiniRust.expr
 (* However, generally, we will have a type provided by the context that may
    necessitate the insertion of conversions *)
 and translate_expr_with_type (env: env) (e: Ast.expr) (t_ret: MiniRust.typ): env * MiniRust.expr =
+  let rec erase_lifetime_info (t: MiniRust.typ) = match t with
```
msprotz, line 592: can you do this with a visitor? it might be more future-proof

R1kM: Fixed in [0007e91](https://github.com/FStarLang/karamel/pull/441/commits/0007e91fed9d8c7b131e432e3a636c4638ffc3c7). As discussed in person, this required a few tweaks to the MiniRust AST to generate visitors for typ.

msprotz (approved):

PR #446, opened 2024-07-14
(RFC) Add a test for mul_underspec

mtzguido:
Hi Jonathan, this PR adds a test for `mul_underspec` to test the patch in https://github.com/FStarLang/FStar/pull/3349.

I had to use this `irreducible` four definition to prevent the normalizer from simpliyfing the operation, so we actually try to extract it. Not sure if you have a better way?

I don't know if this is of much value as-is, maybe we can beef it up with other operations/widths too. For that I think `main` would look like
```fstar
if not (test1 ()) then 1l else
if not (test2 ()) then 1l else
if not (test3 ()) then 1l else
...
```
or something similar.

### DIFF
--- a/test/.hints/Underspec.fst.hints
+++ b/test/.hints/Underspec.fst.hints
@@ -1,0 +1,42 @@
+[
+  "B/É¸\u0002Ñ\u000fÁF\u0017¬—≤zê\u0006",
+  [
+    [
+      "Underspec.test1",
+      1,
+      2,
+      1,
+      [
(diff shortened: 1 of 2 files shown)


msprotz:
I usually use an effectful function that returns a constant to prevent optimizations. But this is fine too! Thanks, this looks good to me, you can go ahead and merge. Thanks so much.

mtzguido:
Thanks! Sorry, I should have removed the (RFC) before merging...

PR #462, opened 2024-08-20
Do not generate (); in Rust backends

R1kM:
This PR does three things:
* It slightly refactors Rust simplification passes into a single module, split into passes on Ast and passes on MiniRust
* It does not emit unit statements when on the left-hand side of a sequence anymore
* It removes trailing units in function bodies


lib/PrintMiniRust.ml
```diff
@@ -236,6 +236,9 @@ and print_expression_with_block env (e: expr): document =
 
 and print_statements env (e: expr): document =
   match e with
+  | Let ({ typ = Unit; _ }, Unit, e2) ->
```
msprotz, line 239: do you understand why this is not caught by the previous passes over the `Ast`?

lib/SimplifyRust.ml
```diff
@@ -2,13 +2,86 @@
...
   method! visit_EPopFrame _ = EUnit
 end
 
-let simplify files =
+let simplify_ast files =
   let files = remove_push_pop_frame#visit_files () files in
   files
+
+
+(* Transformations occuring on the MiniRust AST, after translation and mutability inference *)
```
msprotz, line 19: do you mind putting those in OptimizeMiniRust? this is the file that has optimizations, for Rust, over the MiniRust AST, while SimplifyRust has optimizations, for Rust, on the Ast AST.

msprotz (approved):
modulo moving stuff around a little bit

lib/PrintMiniRust.ml
R1kM: Some of them are introduced by the translation of EBufFree in the Ast to MiniRust pass, which thus occurs after the Ast simplification passes. I'm not sure this is the case for all, but that at least gives some unhandled cases.

lib/SimplifyRust.ml
R1kM: Done in [9465c7d](https://github.com/FStarLang/karamel/pull/462/commits/9465c7d566ab7c55b0c9491888dfb43f7c5acc11)

lib/PrintMiniRust.ml
msprotz: cool, if you can, a comment would help -- so many other phases *already* do this earlier that it's a little surprising to see this here; it makes perfect sense! but a comment would help a bit

msprotz (approved):

lib/PrintMiniRust.ml
R1kM: Good point, this is done in [a62fa60](https://github.com/FStarLang/karamel/pull/463/commits/a62fa609c9309d4be2c2eb9a44c32446245fbc6a)

PR #469, opened 2024-08-22
Introduce -faggressive-inlining

mtzguido:
Hi Jonathan, thanks for the tip yesterday about the inlining phase! It does make our code look much nicer.

In fact, I was thinking I would like this to be always done regardless of name and without requiring any attributes, so thought about adding this option, do you think it would be useful for others?

I could also add an attribute-based mechanism like you suggested if that's useful. If so, and if this option isn't desirable, maybe I can just have Pulse always add the attribute.

### DIFF
--- a/lib/Options.ml
+++ b/lib/Options.ml
@@ -75,6 +75,7 @@ let microsoft = ref false
 let extern_c = ref false
 let short_names = ref true
 let cxx_compat = ref false
+let aggressive_inlining = ref false
 
 let extract_uints = ref false
 let builtin_uint128 = ref false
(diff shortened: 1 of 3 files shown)


msprotz:
Personally, I'd rather have an attribute, since currently we're doing a bunch of hacks in HACL-rs to name some variables `uu_` to avoid intermediary let-bindings that cause move-outs in Rust (see https://github.com/hacl-star/hacl-star/blob/afromher_rs/code/streaming/Hacl.Streaming.Functor.fst#L270 for instance). It would be cleaner if we could mark these let-bindings as `[@inline]` with a descriptive name (instead of uu_).

But I also appreciate that the "clean" solution is probably a little bit of work to propagate, so I'm fine with the flag too. Feel free to merge is that makes a significant difference for you (with the understanding that none of my projects will use this flag, so you'll have to keep an eye on it to make sure it doesn't regress).

mtzguido:
Sounds good, let me try that and get back.

mtzguido:
This was superseded by #470. I don't really need the flag anymore, hence closing, but I rebased it just in case it becomes useful at a point. We could also consider an F* flag to always add the attribute.

PR #475, opened 2024-08-30
Prevent move-out on pointer holding structs

R1kM:
As described in the title. There currently is some special-casing to avoid move-out when, e.g., accessing an array of non-Copy elements.
This PR extends what is considered as a non-Copy element to also consider pointer holding structs.

### DIFF
--- a/lib/AstToMiniRust.ml
+++ b/lib/AstToMiniRust.ml
@@ -828,7 +828,7 @@ and translate_expr_with_type (env: env) (e: Ast.expr) (t_ret: MiniRust.typ): env
       let t = translate_type env b.typ in
       let is_owned_struct =
         match b.typ with
-        | TQualified lid when Idents.LidSet.mem lid env.heap_structs -> true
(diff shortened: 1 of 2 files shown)


msprotz:
I take it pointer_holding_structs only contains *mutable* pointers? I guess we don't know at that stage so we have to be conservative...

R1kM:
Actually, since we do not derive the Copy or Clone traits for structures, the check should probably be more restrictive, and check for any struct type.

For instance, despite using a struct that does not contain pointers, the following Rust code does not compile

```rust
struct T {
    x: u32,
    y: u32
}

fn g(s: &[T]) {
    let x: T = s[0];
}
```

Coming back to the a simplified version of the HACL-rs that inspired this PR, the following code would indeed work without the additional reference added by karamel, while taking a mutable pointer in struct `T` would lead to a compilation failure.

```rust
#[derive(Copy, Clone)]
struct T <'a> {
    x: u32,
    y: &'a [u32]
}

fn g(s: &[T]) {
    let x: T = s[0];
}
```

However, adding support for Clone/Copy on generated struct types does not seem needed at the moment, and this conservative pattern is sufficient for compilation. I would suggest to revisit only if needed.

PR #479, opened 2024-09-09
Simplify hoisting logic

msprotz:


msprotz:
This case was obviously suboptimal -- found out while doing code quality improvements for kyber.

PR #483, opened 2024-09-13
Better handling of comments

msprotz:
The goal of this PR is to move the handling of in-expression comments from dedicated AST nodes to meta-level attributes on nodes. The reason for this is that the EComment node potentially trips a lot of transformation passes that perform large pattern-matches on AST nodes that may start failing if an EComment node is found in the wrong place.

Rather than add an (expensive) strip_comments function everywhere, I chose instead to move the comments to meta-level attributes on nodes themselves, which I believe makes things much easier to handle.

The problem is that this, in turn, drops comments unless I audit every single pass to make sure it doesn't fill out an empty default meta... so I'm not sure this is worth merging..


msprotz:
After carefully reviewing our options here, I think this is the best bet: I'd rather get an annoying, but not life-threatening issue where comments are dropped, rather than a semantically meaningful phase that silently stops operating because of an EComment node in the middle of a term that prevents a pattern from matching.

@tahina-pro you are the biggest user of the "inline comment" feature, which I attested still works (strengthened test forthcoming) -- unless I hear from you within the next couple days, I'll just merge this and assume that everparse is not affected, thank you

PR #485, opened 2024-09-18
specify deprecated attribute for Oracle Studio compiler

vladak:
Another change for Oracle Studio compiler - this time adding the support for the deprecated attribute. Again, assuming fairly modern version of the compiler (>=12.x).


msprotz:
Thank you!

PR #487, opened 2024-09-20
Rust backend: allow const_item_mutation warning, simplify deref_addrof pattern

R1kM:
With the recent addition of mutability inference, accesses to top-level constants do not require mutable borrows anymore, which was the cause of the const_item_mutation warning.

Additionally, this PR adds a micropass to simplify the `deref_addrof pattern`, e.g., `*&[]`


msprotz (approved):

msprotz:
Not sure why CI failed here

PR #494, opened 2024-10-22
Change the criterion for KRML_IGNORE vs KRML_MAYBE_UNUSED_VAR

msprotz:
before, when discovering that `x` is unused in `let x = e1 in e2` , we would unconditionally push `KRML_IGNORE` down `e1` (possibly beneath multiple layers of let-bindings)

now, instead, we either do `let x = KRML_IGNORE(e1) in e2` is `e1` is a value, otherwise, we do `let x = e1 in KRML_MAYBE_UNUSED_VAR(x); e2`


msprotz:
(waiting to see the result for code quality in HACL*)

msprotz:
(Actually generates slightly better code for HACL*)

PR #497, opened 2024-11-21
Temporary CI fixes due to FStarLang/FStar#3615

tahina-pro:
Since Ubuntu 23.10 is EOL, and 24.04 does not work with F* CI, the latter was reverted to Ubuntu 22.04, see FStarLang/FStar#3615 

This PR adapts to this change.


msprotz:
Thanks Tahina!

PR #503, opened 2024-12-06
Makefiles: fix KRML_HOME to work with Cygwin

tahina-pro:
Some Makefiles were defining `KRML_HOME` with Cygwin paths. This PR fixes them to native Windows paths.

### DIFF
--- a/krmllib/Makefile
+++ b/krmllib/Makefile
@@ -114,9 +114,14 @@ $(MINI_DIR):
 $(GENERIC_DIR):
 	mkdir -p $@
 
+local_krml_home := $(realpath $(CURDIR)/..)
+ifeq ($(OS),Windows_NT)
+  local_krml_home := $(shell cygpath -m "$(local_krml_home)")
+endif
+
 # Everything in the universe
 $(GENERIC_DIR)/Makefile.include: $(ALL_KRML_FILES) | $(GENERIC_DIR) $(wildcard c/*.c) $(wildcard c/*.h) ../_build/default/src/Karamel.exe
-	KRML_HOME=$(shell pwd)/.. \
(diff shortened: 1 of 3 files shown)


msprotz (approved):
CURDIR is good but only supported by GNU Make, so if this Makefile doesn't do it already, I would suggest aborting in case someone is running e.g. BSD make

tahina-pro:
Thanks Jonathan for your heads up. What about `$(shell pwd)`, as is currently done in `master`? Do any BSD variants of `make` support it? If so, then I'd rather switch back to `$(shell pwd)`

(For the record, with this PR as it is now, `./everest make` and `./everest test` successfully work on Windows+Cygwin with official opam 2.3 (and OCaml 5.2.1): https://github.com/project-everest/everest/actions/runs/12239757622 .)

tahina-pro:
Thanks Jonathan!
Yes, I can confirm `$(shell pwd)` works well with `make -C dir`.

I didn't find any reliable way to check for GNU make other than renaming the Makefile into GNUmakefile, per https://www.gnu.org/software/make/manual/html_node/Makefile-Names.html : 
> By default, when make looks for the makefile, it tries the following names, in order: GNUmakefile, makefile and Makefile.


Alternatively, the check in the HACL* Makefile at https://github.com/hacl-star/hacl-star/blob/b228390cd45329996e80380135d17410bc23e4a7/Makefile#L46
can rule out OSX Make. Would that be enough?

msprotz:
I think it's good as it is then. No need to waste your precious time with any of these details, you make no more assumptions than we already do in the existing build and other parts of Everest. Thank you for double-checking!

tahina-pro:
Thanks a lot Jonathan for merging!

PR #504, opened 2024-12-13
Use _mm_malloc() for KRML_ALIGNED_MALLOC on macOS <10.15

aeiouaeiouaeiouaeiouaeiouaeiou:
Discovered when [testing the Python 3.14 beta](https://github.com/macports/macports-ports/pull/26974) on older versions of macOS.

PR #505, opened 2024-12-13
Add LEGACY_MACOS macros, use _mm_free() for KRML_ALIGNED_FREE on macOS <10.15

aeiouaeiouaeiouaeiouaeiouaeiou:
I didn't expect #504 to be merged quickly.
Since we are using `mm_malloc.h`, in this case it is better to apply `_mm_free()` to `KRML_ALIGNED_FREE`.

### DIFF
--- a/include/krml/internal/target.h
+++ b/include/krml/internal/target.h
@@ -19,9 +19,18 @@
 #  define inline __inline__
 #endif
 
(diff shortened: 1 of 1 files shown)


msprotz (approved):

msprotz:
Super, thank you! Curious why this wasn't caught before when I did my Python regression testing, but maybe the CI wasn't as extensive there as I thought? In any case, thanks for taking care of this

PR #507, opened 2024-12-20
Add support internally for choosing the value of enum cases

msprotz:
Essentially, adds support for `typedef enum t = { Foo = 42, ... }` in the internal AST, with the understanding that this can come out as a C11 enum (via a command-line flag) or the usual series of `#define`s, and each will assign correct numerical values to each enum case.

For now, this assumes that either all enum cases have an assigned value, or that none have an assigned value. (This could be easily fixed).

CC @tahina-pro who might find this useful for everparse: hit me up if you want more details!


msprotz:
```
#23 5.498 [2024-12-20 18:33:41] [STDERR]  - Z3 solver not found.
#23 5.498 [2024-12-20 18:33:41] [STDERR]  - Required version: 4.8.5
#23 5.879 [2024-12-20 18:33:41] [STDERR]  - Please download version 4.8.5 of Z3 from
#23 5.879 [2024-12-20 18:33:41] [STDERR]        https://github.com/Z3Prover/z3/releases
#23 5.879 [2024-12-20 18:33:41] [STDERR]    and install it into your $PATH as 'z3-4.8.5'.
```

msprotz:
@mtzguido FYI

PR #508, opened 2024-12-20
install-deps.sh: install z3 with the F* script

mtzguido:
The standalone build is currently not installing Z3 as it was removed from the OPAM dependencies. Use the F*-provided script to set it up.

--

The original buils for this PR https://github.com/FStarLang/karamel/pull/507 failed due to going through the "standalone" build instead of the "hierarchic". The latter starts from a built docker image where we ran F* CI, that is kept in the buildmachine's state, while the former starts from scratch and is seldom tested. This PR fixes the standalone build to properly install Z3 after https://github.com/FStarLang/FStar/pull/3631


mtzguido:
Sadly I have no way of triggering a standalone build from the actions UI... but I did test it locally by:
```
docker build -f .docker/standalone.Dockerfile .
```

msprotz:
Thanks Guido!

PR #510, opened 2024-12-25
change z3 version to 4.13.3

karthikbhargavan:
I am sure there may be a better place to put the `--z3version` but with this change, KaRaMeL works with the latest F* and Z3 4.13.3

### DIFF
--- a/krmllib/Makefile
+++ b/krmllib/Makefile
@@ -84,7 +84,7 @@ obj:
 	mkdir $@
 
 %.checked: | obj
-	$(FSTAR) $(OTHERFLAGS) $< && \
+	$(FSTAR)  --z3version 4.13.3 $(OTHERFLAGS) $< && \
 	touch $@
 
 verify-all: $(ALL_CHECKED_FILES)


msprotz:
@mtzguido can you remind me what's up with using the most recent z3 version? 

what I'd like to understand is what happens to people like @R1kM and I for whom z3 is z3 4.8.5 and who presumably don't have z3 4.13.3 installed -- can everything be handled with `./everest z3`? just want to make sure before merging this in!

prosecco:
warning: valecrypt does not seem to like 4.13.3

mtzguido:
Yes running `./everest z3` should install both versions as `z3-4.8.5` and `z3-4.13.3` so F* can find them automatically. https://github.com/project-everest/everest/pull/119

msprotz:
@karthikbhargavan can you recall if you upgraded something else in your setup then? could it be that you now have `z3` resolve to 4.13.3 instead of the old 4.8.5? F* expects 4.8.5 by default

mtzguido:
(btw, I think karamel should work as it is now with 4.8.5 as long as F* can find it, but the idea is to gradually move to 4.13.3, so I'd be happy seeing this PR merged if it works for everyone)

msprotz:
yeah I'm roughly in favor of merging (modulo putting the flag directly in $(FSTAR)), with the caveat that some users may not realize that the right thing to do is to create aliases, and thus will end up with z3 being 4.13.3 which will then make their build of HACL* fail (which is what I suspect happened to Karthik)

maybe the error message should clearly say "please create an alias" (if it doesn't already?)

mtzguido:
The error should be something like this:
```

mtzguido:
With the new F* build we are now publishing nightly builds of F* for Linux and Mac https://github.com/FStarLang/FStar-nightly/releases. The packages include both Z3 versions in a location that can be found by F* (`$PREFIX/lib/fstar/z3-VER/...`), so they don't need to be added to the user's PATH, just call `fstar.exe` and it will figure it out. (Releases will do the same.)

Maybe this will make it easier for everyone? There is also a script in the F* repo to download and setup a nightly tarball instead of building (`.scripts/setup_nightly.sh`). I could also make the proposed new CI (#515) test against the nightly build.

mtzguido:
F* will soon update to use 4.13.3 by default, see [this PR](https://github.com/FStarLang/FStar/pull/3880). Karamel is working fine with that change. This patch won't be needed when that happens, so I think we can close this.

PR #512, opened 2025-01-04
krml/Makefiles: unifying logic for detecting F*

mtzguido:
Mostly preparing for using F* outside of a repo, i.e. without an FSTAR_HOME. The Makefiles and krml will now only locate fstar.exe, and call it to find its library and OCaml files. To find fstar.exe, the process is:
- First try to read the environment variable FSTAR_EXE.
- If unset, try FSTAR_HOME/bin/fstar.exe (for compatibility)
- If FSTAR_HOME is unset, check the PATH.

(I'm proposing other projects should use that logic too.. does it make sense? Steel is already doing via this same `locate_fstar.mk`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,22 +22,11 @@ endif
 
 all: minimal krmllib
 
-ifdef FSTAR_HOME
-  OCAMLPATH:=$(OCAMLPATH)$(OCAMLPATH_SEP)$(FSTAR_HOME)/lib
-else
-  FSTAR_EXE=$(shell which fstar.exe)
-  ifneq ($(FSTAR_EXE),)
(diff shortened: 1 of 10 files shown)


gebner:
> I'm proposing other projects should use that logic too.. does it make sense?

Personally I think it's more complicated than necessary.  In virtually all cases you want to have `fstar.exe` in your PATH anyhow.  So I'd skip step 1 and 2 and just call `fstar.exe` directly.

mtzguido:
Posting an update soon to remove all F*-finding logic from karamel, and using `FSTAR_EXE ?= fstar.exe` in the makefiles.

PR #515, opened 2025-01-08
Github actions CI

mtzguido:
This changes the CI to be completely cloud hosted and not rely on state in the runner machine. I think all features are retained and the .docker directory can be blasted away. Except for Slack messages, I didn't port that chunk as it will not work over multiple jobs, and it's probably better if we find something canned instead of retaining the old code. (Plus maybe move to Zulip?)

It does use a base container hosted in Dockerhub, at least for now. It'd be nicer to keep a container description in the repo and build on-demand.

Note that soon the buildmachine will stop building F* images, so the old workflow will start going through the 'standalone' path and be slower.

Example run: https://github.com/mtzguido/karamel/actions/runs/12663210092


msprotz:
> It does use a base container hosted in Dockerhub, at least for now. It'd be nicer to keep a container description in the repo and build on-demand.

Who is in charge of regenerating the base container? Is this something official that the F* team commits to maintaining? Should we switch other projects like HACL* to it?

mtzguido:
> > It does use a base container hosted in Dockerhub, at least for now. It'd be nicer to keep a container description in the repo and build on-demand.
> 
> Who is in charge of regenerating the base container? Is this something official that the F* team commits to maintaining? Should we switch other projects like HACL* to it?

Currently I'm not regenerating it at all (it's described in `.docker/dev-base.Dockerfile` in F*). It was written in support of check-world, and contains basically all dependencies of all the projects in it. Unsure if we should give it official status... I would try instead to phase this thing out.

I think the right thing to do is just set the environment up with specific actions instead of rolling our own container, and write a `setup-fstar` action. I can give that a try and replace this PR. It would also make it really easy to do a build (maybe just `minimal`) on several OCaml versions, if that's of interest.

msprotz (approved):
looks great -- thanks

mtzguido:
Merging then. We could add a Nix job too if someone writes it up. Also, let me know if you'd like to test several OCaml versions at once, this is pretty easy to do.

PR #516, opened 2025-01-08
No FSTAR_HOME

mtzguido:
With this PR krml will no longer do anything fancy to find an "FSTAR_HOME" or the F* executable: it will use the PATH. It can be overriden with the new `-fstar` option though. The library can be located by just asking F*.

All the makefiles set `FSTAR_EXE ?= fstar.exe` and pass `-fstar $(FSTAR_EXE)` to krml, so one can set FSTAR_EXE instead of placing F* in the PATH. But this is a convention of the Makefiles.


msprotz (approved):
Super happy about this change. @nadrieril any thoughts on the nix changes?

Nadrieril:
Looks good to me

msprotz:
Guido and I talked extensively about this and I just did one last round of review, this is good for me

PR #517, opened 2025-01-08
Fix Rust cosmetic pass removing trailing units

R1kM:
Slight refinement of the nanopass removing trailing units in the Rust backend.
Consider the following code
```rust
let x: u32 = 0;
()
```
With the previous version, the nanopass would have rewritten this code into `0`, leading to typechecking failures in Rust.

This PR tweaks the pass to only perform the rewriting when the previous expression has type unit.


msprotz:
Alternatively, you could rewrite into  `let _x: u32 = 0;` because you know for sure that it is unused. (If it doesn't already start with _).

I'm fine with either. Thanksè

PR #523, opened 2025-01-16
Removing hints

nikswamy:
Following F* PRs https://github.com/FStarLang/FStar/pull/3440 and https://github.com/FStarLang/FStar/pull/3680

This PR removes all hints files from the Karamel repo and updates the Makefiles to use the `--ext context_pruning` flag instead.

With this flag, F* prunes away SMT assertions that cannot be triggered by symbols in the query, making hints/unsat-core replay mostly unnecessary.

We have been using this feature in the F*, Pulse, EverParse, and other related repos for a few months now, and recently removed all hints from the F* repo.

Could this be a simplification for Karamel too? If so, we could also remove the hint-refresh CI jobs.


msprotz (approved):
Great. Hints have always been a band-aid and I'll be happy to see them go. A few thoughts:
1. there is no hints-refresh job on this repo, no need to worry about that
2. the proofs for these files were never substantial, and we never reached the size (like in HACL*) where the files would start triggering unreliable z3 behavior, so hints were never crucial for verification
3. speaking of HACL*, does the repo verify without hints and with the new option? if so, I'd be happy to take a PR there

thanks!

PR #525, opened 2025-01-30
Revise dependency analysis to be more precise

msprotz:
The dependency analysis was pretty coarse, and only represented dependencies from one "file" to either regular or internal headers. It turns out that the dependencies vary quite widely between that file's public header, internal header, or C implementation.

This was pointed out here: https://github.com/hacl-star/hacl-star/pull/1007#discussion_r1871934044

This revamps the analysis to now distinguish dependencies between { h file, internal/h file, c file } to { h file, internal/h file }.

The net result is a considerably reduced number of #includes in public headers.

### DIFF
--- a/.github/workflows/linux-x64-hierarchic.yaml
+++ b/.github/workflows/linux-x64-hierarchic.yaml
@@ -16,7 +16,7 @@ jobs:
         run: |
           echo "CI_INITIAL_TIMESTAMP=$(date '+%s')" >> $GITHUB_ENV
       - name: Check out repo        
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


msprotz:
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

msprotz:
@mtzguido do we need the other PR merged for this to work? if so, happy to approve it (or for you to merge it)... thanks!

mtzguido:
I think this simple patch should do it. You had very valid concerns for the other CI so maybe it can wait a bit.

msprotz:
Thanks!

PR #531, opened 2025-02-10
Rework C compiler detection

### LINKED ISSUE #529: cc autodetection?
This snippet
https://github.com/FStarLang/karamel/blob/c92ffed8f3dc381b9306ac89db356711ecf194c2/lib/Driver.ml#L342-L356
causes krml in my system to use a cross-compiler, when I wasn't expecting it to (so `make test` will fail, as will other projects). I have `x86_64-w64-mingw32-gcc` in the PATH, and also `gcc-{12,13,14}` (but not 10 or lower).

I could add some more numbers to this list to make it pick my local gcc... but I was wondering what the purpose of this code even is. I would have expected karamel to just call `cc` (be it gcc or clang according to the user's setup), and maybe be overridable by the `-cc option` or even the standard `CC` environment variable.

mtzguido:
This patch makes karamel pick, by default, 'cc' from the path as the C compiler. This command can be overriden with the -cc option (taking the name of a program in the PATH, or a full path to an executable), or the CC environment variable (idem).

Since karamel passes specific options to different compilers, there is now logic to detect the flavor of compiler that karamel is talking to (e.g. by looking at `cc --version`).

*MISSING*: MSVC autodetection, and fixing the wrapper script. I don't have enough experience using it to know if we can tweak this script so as to only point it to an executable.

Also I think I should add a way to override the flavor in case autodetection fails.

(If you feel this change is more trouble than it's worth I'm ok dropping it.)

Fixes #529

### DIFF
--- a/lib/Driver.ml
+++ b/lib/Driver.ml
@@ -6,34 +6,40 @@
 
 open Warn
 
-(* Abstracting over - (dash) for msvc. vs. gcc-like. *)
+(** These three filled in by [detect_cc] and others *)
+let cc = ref ""
+let cc_args = ref []
+let cc_flavor = ref Options.Generic
+
+(* Abstracting over - (dash) for msvc. vs. gcc-like. This module
+ reads cc_flavor. *)
 module Dash = struct
   let i dir =
-    if !Options.cc = "msvc" then
+    if !cc_flavor = MSVC then
       [ "/I"; dir ]
     else
       [ "-I"; dir ]
 
   let c file =
-    if !Options.cc = "msvc" then
+    if !cc_flavor = MSVC then
       [ "/c"; file ]
     else
       [ "-c"; file ]
 
   let d opt =
-    if !Options.cc = "msvc" then
+    if !cc_flavor = MSVC then
       "/D" ^ opt
     else
       "-D" ^ opt
 
   let o_obj file =
-    if !Options.cc = "msvc" then
+    if !cc_flavor = MSVC then
       [ "/Fo:"; file ]
     else
       [ "-o"; file ]
 
   let o_exe file =
-    if !Options.cc = "msvc" then
+    if !cc_flavor = MSVC then
       [ "/Fe:"; file ]
     else
       [ "-o"; file ]
@@ -43,6 +49,8 @@ end
 module P = Process
 (* Note: don't open [Process], otherwise any reference to [Output] will be
  * understood as a cyclic dependency on our own [Output] module. *)
+
+let cc_flavor_callback = ref (fun (_ : Options.compiler_flavor) -> ())
 
 (** These three variables filled in by [detect_fstar] *)
 let fstar = Options.fstar
@@ -56,10 +64,6 @@ let runtime_dir = ref ""
 let include_dir = ref ""
 let misc_dir = ref ""
 let krml_rev = ref "<unknown>"
-
-(** These two filled in by [detect_gcc] and others *)
-let cc = ref ""
-let cc_args = ref []
 
 (** The base tools *)
 let readlink = ref ""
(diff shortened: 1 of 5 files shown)


msprotz:
I'm generally in favor. I'd be surprised if anyone ran this in a windows command shell expecting krml to pick up on msvc -- if there are such people, we'll hear about them, but I doubt it.

msprotz:
Also, the bat script is strange, because most people would open up a VS command prompt *then* invoke krml (meaning the right includes would be set in the path). I think that's a remnant of an old era where krml was run from a cygwin prompt and was expected to source the right environment variables itself.

mtzguido:
Sounds good, thanks. Maybe then I can remove the script altogether and make
it uniform to the other flavors. I'll update today.

mtzguido:
Updated. Just calling `cl.exe` seems to kind of work... except that it doesn't find standard headers:
```
$ CC=cl.exe make test
[...]
../_build/default/src/Karamel.exe -fstar fstar.exe  -warn-error @4 -verbose -skip-makefiles  -tmpdir .output/Abbrev.out -no-prefix Abbrev \
  -o .output/Abbrev.exe [...] -bundle Abbrev=WindowsHack,\*
✔ [Monomorphization] ⏱️ 58ms
✔ [Inlining] ⏱️ 7ms
✔ [Pattern matches compilation] ⏱️ 7ms
✔ [Structs + Simplify 2] ⏱️ 8ms
✔ [Drop] ⏱️ <1ms
✔ [AstToCStar] ⏱️ 2ms
✔ [CStarToC] ⏱️ <1ms
⚙ KaRaMeL auto-detecting tools. Here's what we found:
readlink is: readlink
KaRaMeL called via: ../_build/default/src/Karamel.exe
KaRaMeL home is: /home/guido/r/karamel
✔ [PrettyPrinting] ⏱️ 4ms
KaRaMeL: wrote out .c files for Abbrev
KaRaMeL: wrote out .h files for Abbrev
⚙ KaRaMeL will drive the C compiler. Here's what we found:
cc is: cl.exe (= /mnt/c/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.42.34433/bin/Hostx64/x64/cl.exe)
cc flavor is: msvc
cl.exe options are: -I /home/guido/r/karamel/krmllib/dist/minimal -I /home/guido/r/karamel/krmllib -I /home/guido/r/karamel/include -I .output/Abbrev.out
⚡ Generating object files
cl.exe -I /home/guido/r/karamel/krmllib/dist/minimal -I /home/guido/r/karamel/krmllib -I /home/guido/r/karamel/include -I .output/Abbrev.out -c .output/Abbrev.out/Abbrev.c -o .output/Abbrev.out/Abbrev.obj
✘ [CC,.output/Abbrev.out/Abbrev.c]
Microsoft (R) C/C++ Optimizing Compiler Version 19.42.34436 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

cl : Command line warning D9035 : option 'o' has been deprecated and will be removed in a future release

Abbrev.c
\\wsl.localhost\Ubuntu\home\guido\r\karamel\include\krml/internal/target.h(7): fatal error C1083: Cannot open include file: 'assert.h': No such file or directory

Warning 3: run_or_warn: the following command failed:
cl.exe -I /home/guido/r/karamel/krmllib/dist/minimal -I /home/guido/r/karamel/krmllib -I /home/guido/r/karamel/include -I .output/Abbrev.out -c .output/Abbrev.out/Abbrev.c -o .output/Abbrev.out/Abbrev.obj
Warning 3 is fatal, exiting.
make[1]: *** [Makefile:131: .output/Abbrev.exe] Error 255
make[1]: Leaving directory '/home/guido/r/karamel/test'
make: *** [Makefile:57: test] Error 2
```
but these are in fact not installed in my system, so maybe I got a partly broken MSVC. There also seems to be no `vcvarsall` script for the 2022 edition I installed, maybe the need for that is gone? (Also note the warning about /o surprisingly being deprecated!?) Would be useful if someone has a working MSVC and wants to test.. but maybe no one is doing that.

(This was in WSL, same thing happens from a plain cmd shell.)

Also added an override for the flavor detection and fixed a few quirks.

If you want to merge this I'll clean up the history and mark ready.

msprotz:
Yeah I think that's the reason d'être of the script -- it calls vcvarsall which sets the right environment variables to help cl find the includes (I think)

mtzguido:
You're right, somehow I got mixed up and couldn't find the varsall script in my installation, but I see it now and it does make cl work. However I don't have any of the variables that the script looked for (`%VS150COMNTOOLS%`, etc) set.

I suppose calling this script *and then* running krml would work fine, but no idea how the environments would interact if krml is ran from a cygwin shell.

Also apparently starting with Visual Studio 2017 the situation is different, the variables are not set and one needs to (somehow find and then) call a shell wrapper script. https://help.appveyor.com/discussions/kb/48-vs2017.

So... a big mess. Any thoughts? Maybe one way around this is specially recognizing when cc equals `msvc`, and just using all the old logic there. I don't think it will work in VS 2017 and upwards though.

msprotz:
Yeah now you're supposed to use vswhere for that purpose: https://github.com/microsoft/vswhere

Maybe that can replace the old bat script?

mtzguido:
Updated to special-case `-cc msvc` (or `CC=msvc`) and call the wrapper script if so. 

Also added a case in that script to call vswhere if available, and that seems to work locally. Separate PR since it's independent from this https://github.com/FStarLang/karamel/pull/535.

mtzguido:
Updated, thanks!

PR #533, opened 2025-02-11
Support Assumed Declaration in MiniRust translation

R1kM:
PR in support of scylla, to integrate external (assumed) declarations when transpiling from C to Rust.

This PR translates `DExternal` declarations in Ast to a new `Assumed` node in MiniRust.
In particular, the translation preserves the mutability that was specified by the external declaration, and uses it for the mutability inference in `OptimizeMiniRust`. This behavior is overriden for external, "builtin" definitions.

Assumed definitions are then filtered and not extracted to Rust: Rust clients are expected to provide them directly.

The extraction of hacl-rs from Low* remains the same with this PR.


msprotz (approved):
Looks good to me! Thanks.

PR #536, opened 2025-02-14
Simplify Makefiles after F* changes

mtzguido:
Mostly:
1. For extracting, we can just pass the path of the checked file to
   F* instead of doing the hacky substitution and rely on F* to find the
   source module.
2. no need to pass --extract when we want to extract a single file
   (except for --codegen krml, separate PR incoming for that).
3. Using -o instead of (the now deprecated) --output_deps_to and
   --krmloutput.
4. Using -c instead of --cache_checked_modules, and only passing it
   when we are actually trying to create a checked file.

Also added some explicit -o flags to make sure F* writes to the desired
target.

(Previous text:

Mostly no need to pass --extract when we want to extract a single file, and we can just pass the path of the checked file to F* instead of doing the substitution.

Also added some explicit --krmloutput flags to make sure F* writes to the desired target.)

### DIFF
--- a/book/tutorial/Makefile
+++ b/book/tutorial/Makefile
@@ -46,11 +46,10 @@ FSTAR_EXTRACT = --extract 'OCaml:-* +Spec'
 # But also:
 # - --cmi, for cross-module inlining, a must-have for anyone who relies on
 #   inline_for_extraction in the presence of interfaces
-# - --cache_checked_modules to rely on a pre-built ulib and krmllib
-# - --cache_dir, to avoid polluting our generated build artifacts outside o
+# - --cache_dir, to place checked files there
 
 FSTAR_NO_FLAGS = $(FSTAR_EXE) --ext context_pruning \
-  --odir obj --cache_checked_modules $(FSTAR_INCLUDES) --cmi \
+  --odir obj $(FSTAR_INCLUDES) --cmi \
   --already_cached 'Prims FStar LowStar C Spec.Loops TestLib WasmSupport' --warn_error '+241@247+285' \
   --cache_dir obj
 
@@ -72,7 +71,7 @@ FSTAR_ROOTS = $(wildcard $(addsuffix /*.fsti,$(SOURCE_DIRS))) \
 # mtime of the moved files.
 ifndef MAKE_RESTARTS
 .depend: .FORCE
-	$(FSTAR_NO_FLAGS) --dep full $(notdir $(FSTAR_ROOTS)) $(FSTAR_EXTRACT) --output_deps_to $@
+	$(FSTAR_NO_FLAGS) --dep full $(notdir $(FSTAR_ROOTS)) $(FSTAR_EXTRACT) -o $@
 
 .PHONY: .FORCE
 .FORCE:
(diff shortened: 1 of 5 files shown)


mtzguido:
Wondering if
1- We should just add `-o` option that subsumes `--krmloutput` and `--output_deps_to`, and could even be used for the location of the generated checked file.
2- Is the default behavior for krml (extracting all modules) desirable, or should we make it just extract the module given in the command line, and possibly override with `--extract *`? This would remove the remaining $(subst..) calls.

msprotz:
1- yeah I was just about to suggest the same thing it's weird to have specific options for specific purposes -- most tools use -o regardless of the type of file produced (e.g. gcc -E, gcc -MD, gcc -c, gcc -shared, produce .c, .d, .o, .so but all are controlled by -o)
2- it's historical, and makes it easy to have krml invoke F* to produce a single out.krml file, but it's really bad because there are differences in behavior, and the mode where F* processes multiple files all at once is not really supported anymore -- I'd be in favor of getting rid of it, but then krml would have to output build stubs and defer extraction and compilation to make, which I'm 100% happy about, it's just a fair amount of implementation work

msprotz:
I guess? Like krml would invoke F* with `fstar --extract '*' MyFile.fst -o out.krml`? I think the WASM backend needs WasmSupport.fst passed to F* too, and perhaps even FStar.UInt128 passed to F* too -- I don't remember. I don't really use this mode either.

mtzguido:
Yeah exactly, that should make F* behave just as it does now. I'll give it a go and update here, it'd be nice to make this consistent.

mtzguido:
Updated after implementing `-o` in F*. For changing the behavior of krml extraction, I'll post a separate PR since I expect I may need to go through many projects to get it to work, this bit was backwards-compatible.

krmllib/Makefile
```diff
@@ -77,15 +77,15 @@ obj:
...
 %.checked: | obj
-	$(FSTAR) $(OTHERFLAGS) $< && \
+	$(FSTAR) $(OTHERFLAGS) -c $< -o $@ && \
 	touch $@
 
 verify-all: $(ALL_CHECKED_FILES)
 
 $(EXTRACT_DIR)/%.krml: | .depend
-	$(FSTAR) $(OTHERFLAGS) --codegen krml \
+	$(FSTAR) $< --codegen krml \
```
msprotz, line 86: why drop OTHERFLAGS?

test/sepcomp/a/Makefile
```diff
@@ -35,12 +35,11 @@ include Makefile.include
...
 # - --cmi, for cross-module inlining, a must-have for anyone who relies on
 #   inline_for_extraction in the presence of interfaces
-# - --cache_checked_modules to rely on a pre-built ulib and krmllib
-# - --cache_dir, to avoid polluting our generated build artifacts outside o
+# - --cache_dir, to place checked files there
 
 FSTAR_EXE ?= fstar.exe
 FSTAR_NO_FLAGS = $(FSTAR_EXE) \
-  --odir obj --cache_checked_modules $(FSTAR_INCLUDES) --cmi \
+  --odir obj $(FSTAR_INCLUDES) --cmi \
```
msprotz, line 42: so cache_checked_modules is now the default?

msprotz (commented):
Looks good to me minus my comments. Thanks for this!

book/tutorial/Makefile
```diff
@@ -115,32 +114,22 @@ obj:
...
+	$(FSTAR) --codegen OCaml $< -o $@
 
+# By default, krml extraction in F* will extract every module into a
+# single out.krml file. Make sure to extract the single module we want.
 .PRECIOUS: obj/%.krml
 obj/%.krml:
-	$(FSTAR) $(notdir $(subst .checked,,$<)) --codegen krml \
-	--extract_module $(basename $(notdir $(subst .checked,,$<)))
+	$(FSTAR) $< --codegen krml \
+	--extract_module $(basename $(notdir $(subst .checked,,$<))) \
```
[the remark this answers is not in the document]
mtzguido: --codegen krml still has the default extracting everything (instead of just the module in cmdline). I'll change that soon in a separate PR as it'd be a breaking change for many projects (I think)

test/sepcomp/a/Makefile
mtzguido: We always read checked files by default (this was gated by this option at some time long ago, but not now). This option only control whether to *write* the checked files, so I'm only using it (`-c`) in the rules to generate checked files. This is cleaner and less of a footgun IMO but happy to revert if you prefer.

krmllib/Makefile
mtzguido: whoops that's a mistake, updated

test/sepcomp/a/Makefile
```diff
@@ -106,32 +105,24 @@ obj:
...
-#
-# fstar Bignum.Impl.fst --extract_module BigNum.Impl
-#
-# We use basename because we may also extract krml files from .fsti.checked
-# files (not true for OCaml, we don't extract mlis from fstis).
 .PRECIOUS: obj/%.ml
 obj/%.ml:
-	$(FSTAR) $(notdir $(subst .checked,,$<)) --codegen OCaml \
+	$(FSTAR) $< --codegen OCaml \
 	--extract_module $(basename $(notdir $(subst .checked,,$<)))
```
msprotz, line 116: then should this extract_module go? (per your comment)

book/tutorial/Makefile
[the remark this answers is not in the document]
msprotz: ok sounds good -- uniformity is great

test/sepcomp/a/Makefile
msprotz: no this is great thanks for the explanation!

test/sepcomp/a/Makefile
mtzguido: Indeed, updated.

PR #537, opened 2025-02-14
Allow const source buffers for EBufBlit

tahina-pro:
In FStarLang/pulse#326, I am introducing Pulse mutable slices, and I am turning slices into immutable slices. For their C implementation, I am introducing `const` array pointers.
Then, I tried to refine the signature of corresponding memcpy operations so that the source pointer can be `const`. However, Karamel rejects me with:
>  blit requires a non-const pointer

This PR solves this issue by weakening the checks in `lib/Check.ml` to allow immutable source buffers. The destination buffer must still be mutable, though.

Everest green CI: https://github.com/project-everest/everest/actions/runs/13336176953

### DIFF
--- a/lib/Checker.ml
+++ b/lib/Checker.ml
@@ -402,10 +402,9 @@ and check' env t e =
       check_array_index env e3;
       c TUnit
 
-  | EBufBlit (b1, i1, b2, i2, len) ->
-      let t1, c1 = infer_buffer env b1 in
-      check_mut env "blit" c1;
-      check env (TBuf (t1, false)) b2;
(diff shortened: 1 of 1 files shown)


msprotz:
Looks good to me! Thanks so much.

You might want to check other buffer operations like EBufFill to see if they require the same treatment.

tahina-pro:
Thanks Jonathan!
`EBufFill` requires no source buffer, only a destination buffer, and as far as I understand, `Checker.check` still requires it to be non-const, as expected:
https://github.com/FStarLang/karamel/blob/481d45625e8fb31756c343631cedff9efb8493ba/lib/Checker.ml#L398-L401
(and similarly for `infer'`)

PR #542, opened 2025-02-19
Fix long-standing issue with C Abstract Structs

msprotz:
For the longest while we could not enable the usage of CAbstractStruct to hide internal implementation details of the state of the instances of the streaming functor. This was an implementation limitation where the dependency analysis, visibility analysis, and code generation did not know what to do.

Now, the behavior is as follows:
- C abstract structs may be promoted to "internal" visibility, meaning that they now may appear in internal/foo.h
- the visibility and dependency analyses are aware of this behavior, meaning other modules can depend on a C abstract struct
- if a module needs to *see* the definition of an abstract struct, it will include the internal header (not the public one) to preserve the abstraction boundary -- this actually happens in HACL

This was enabled by a recent refactoring of this code that cleans this up significantly.


msprotz:
This is a first step towards option 3 as described in https://github.com/python/cpython/issues/130213#issuecomment-2669451783

msprotz:
A test has broken because a monomorphized instance of a data type has not been marked CAbstractStruct.

I'm kind of on the fence about this: see https://github.com/hacl-star/hacl-star/pull/1025/files for the impact, which is presumably nice, but still needs more work before the usage of libintvector.h can be completely removed from public header.

msprotz:
This work is now complete and the abstraction boundaries in HACL* have been completely overhauled to shield multiplexing modules (e.g. streaming hmac) from having to even *know* about the existence of vectorized types (for vectorized hash implementations).

@tahina-pro are you relying on the behavior of the AbstractStruct annotation? you might want to check this -- thank you!

PR #543, opened 2025-02-20
Honor rename-prefix for *external* Rust declarations

msprotz:
With the current name generation scheme, every file is implicitly subjected to rename-prefix, except for external definitions, which are not emitted. For those, we allow the user to tweak references to these hand-written functions via rename-prefix.


msprotz:
Confirmed no diff on hacl-rs

PR #547, opened 2025-03-05
Improve the Rust extraction of Return nodes

R1kM:
As the translation from Ast to MiniRust is type-directed, translating a Return node needs the return type of the function being currently translated. This PR adds this type to the translation functions (and the mutability inference), and uses it for the EReturn/Return nodes.

It also fixes the translation of a split tree access with a missing `possibly_convert`.


msprotz:
Great thank you!

PR #549, opened 2025-03-06
Support for user-provided annotations about box types in Ast to Rust translation

R1kM:
The Ast to Rust translation relies on several heuristics to determine when pointer types should be translated to boxes.
This PR allows users to specify an additional list of types that *must* be translated to boxes.
The behavior for the F* to Rust extraction is left unchanged: this set is initialized to empty.

### DIFF
--- a/lib/AstToMiniRust.ml
+++ b/lib/AstToMiniRust.ml
@@ -1540,7 +1540,12 @@ let identify_path_components_rev filename =
     components := String.sub filename !start (String.length filename - !start) :: !components;
   !components
 
-let compute_struct_info files =
+(* Compute information about the struct types in the program.
+   In particular, determine whether pointers inside a struct
+   type should be `Box`es or borrows.
+   The [boxed_types] argument corresponds to types explicitly
+   annotated as boxed during C to Rust translation *)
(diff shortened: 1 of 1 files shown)


lib/AstToMiniRust.ml
```diff
@@ -1562,7 +1567,7 @@ let compute_struct_info files =
     method! visit_DFunction _ _ _ _ _ ret_t _ _ _ =
       match ret_t with
       | TQualified lid -> Idents.LidSet.singleton lid
-      | _ -> Idents.LidSet.empty
+      | _ -> boxed_types
```
msprotz, line 1570: seems like it'd be easier to just do `Idents.LidMap.union returned boxed_types`

otherwise looks good thanks!

msprotz (approved):
good modulo my comment, it just makes for a weird visitor otherwise

lib/AstToMiniRust.ml
R1kM: Fixed in [2ab4194](https://github.com/FStarLang/karamel/pull/549/commits/2ab419433ea0d871b6de525c7a3d34d36d4aa872). Thanks!

PR #550, opened 2025-03-06
Actually hash-cons data type tags

msprotz:
So I found out while debugging strange header inclusion issues in HACL* that krml only partially hash-consed data type tags. Concretely, if two monomorphized types `option_t` and `option_u` were in the same file Foo, then they would both rely on `typedef enum { Foo_None, Foo_Some } Foo_option_tags;`. However, if a third `option_v` was placed in file Bar, then this time a different declaration would be emitted, namely, `typedef enum { Bar_None, Bar_Some } Bar_option_tags;`.

This doesn't seem like the end of the world, HOWEVER, in the case of HACL*, and following the recent changes to take into account allocation failures in the streaming functor, a whole bunch of monomorphized instances of `option` started popping up across the codebase. Because of the proliferation of tag types, and combined with the issue above, this meant that enums were inserted in random places, causing cross-dependencies across algorithms.

### DIFF
--- a/lib/DataTypes.ml
+++ b/lib/DataTypes.ml
@@ -239,15 +239,18 @@ let build_scheme_map files =
         | _ ->
             ()
   end)#visit_files () files;
-  map, Hashtbl.create 41
+  map, Hashtbl.create 41, Hashtbl.create 41
 
 (** Second thing: handle the trivial case of a data type definition with only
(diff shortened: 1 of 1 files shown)


msprotz:
This PR actually enforces proper hash-consing of tag types for enums and tagged unions, across all files.

msprotz:
Got a local everest green here

PR #555, opened 2025-03-13
Add a new mode for generating C++17 code

msprotz:
Code generated by krml by default works both in C11 and in C++20 -- the code can be compiled either by a C or a C++ compiler.

For downstream users who do not have C++20, this PR adds a new C++17 mode that generates code compatible with C++17 but *not* with C11. It may be that the code is compatible with C++11 but I have not checked that, so for now, the flag remains C++17.

### DIFF
--- a/lib/C11.ml
+++ b/lib/C11.ml
@@ -63,10 +63,14 @@ and expr =
   | CompoundLiteral of type_name * init list
   | MemberAccess of expr * ident
   | MemberAccessPointer of expr * ident
+  (* The three cases below are NOT in the C grammar *)
   | InlineComment of string * expr * string
   | Type of type_name
   | Stmt of stmt list
-    (** note: last three in the C grammar *)
+  (* Not in the C grammar either; encode a C++ initializer list that appears in expression position
+     so as to rely on an implicit constructor *)
+  | CxxInitializerList of init
+    
 
 (** this is a WILD approximation of the notion of "type name" in C _and_ a hack
  * because there's the invariant that the ident found at the bottom of the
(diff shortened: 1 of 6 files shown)


msprotz:
For now the macro implementations are on the Eurydice side but are intended to move into a krml header eventually.

Most of the problems revolve around creating values that have struct / union types (as opposed to declaring a variable and progressively initializing fields -- krml does have such a pass, but it is currently incorrect in some corner cases, and it generates ugly code, so I'm trying to phase it out).

Basically, in hybrid C11/C++20 mode, krml emits `KRML_CLITERAL(T) { .f1 = e1, ... }`, known as "designed initializer" syntax. This is a C11 / C++20 feature, with the caveat that the syntax is slightly different, and KRML_CLITERAL expands to `(T)` for C11 and `T` for C++20.

If you do not have access to C++20, you can use `T { e1, e2 }` to create a struct of type T, provided the ei are in the same order as the fields fi. (This is an "initializer list", see https://en.cppreference.com/w/cpp/language/aggregate_initialization). But this does not work for unions since you cannot name the case of the union that you intend to construct.

For creating a union, we rely on a rather advanced C++ constructor that uses member pointers and type_identity_t to allow creating a value of the union at the right case, combined with `std::move` to achieve the same effect as a if this was a designated initializer (i.e. no copies).

See https://github.com/AeneasVerif/eurydice/pull/157/files#diff-1af3eebd6ab5727eabc7c3fb1a60f8b6ff2b397bbf934fe8359f5a7fd5296af4R25-R42

PR #568, opened 2025-04-09
Improve mutability inference when assigning to a mutable borrow

R1kM:


R1kM:
The initial C code that inspired this PR (through Scylla) was
```
#include <stdint.h>

void f (uint32_t *a) {
  a[0] = 1;
}

void g (uint32_t *a) {
  f(a);
  a++;
  f(a);
}
```

where `a++` is translated as an assignment, and where one needs to translate the rhs to a mutable borrow to satisfy Rust's type system. Unfortunately, this doesn't help if the last call is commented out: as the analysis is backward, `a` is not in the mutable borrow set by the time `a++` is translated. Closing for now, until a better solution is found (Likely requiring typechecking during mutability inference)

msprotz:
We talked about it and we might need to generate equations for the mutability computation both in a forward and a backward manner.

PR #572, opened 2025-04-11
Improve mutability inference when assigning to a mutable borrow

R1kM:
Follow-up of #568 

Consider the following Rust-like code, provided by Scylla:
```rust
fn f (a: &mut [u32]) {
  a[0] = 1;
}

fn g (a: & [u32]) {
  let _ = f(a);
  let _ = a = &a[1usize..];
}
```

When performing mutability inference of `g`, the current version of karamel will perform a backward analysis, initially starting from an empty set of mutable/mutable borrow variables.
When reaching the assignment `a = a[1usize..]`, as the known sets are empty, inference will not modify the expression, but will add `a` to the set of mutable *variables* (not mutable borrows).
When reaching `f(a)`, as `f` expects a mutable borrow, the inference will add `a` to the set of expected mutable borrows.
After translating the entire body, `a` will therefore be both in the mutable variable and borrow sets, and the function signature will be transformed to `fn g(mut a: &mut [u32])`, leading to a type-checking failure as `a` is assigned an expression with type immutable borrow.

To solve this issue, this PR proposes the following changes:
* Tweak the assignment rule `atom = e`, to force `e` to be a mutable borrow if `atom` is in the mutable borrow set
* When translating a function, initialize the known set with the function parameters that are expected to be mutable borrows
* Perform inference on a function twice after fixpoint computation, to account for the first execution changing the mutability of the function arguments, thus modifying the function body in the second execution.

Note: This solution only covers simple assignments, of the shape `atom = e`, where `atom` is expected to be a mutable borrow. In particular, more complex assignments (e.g., `atom.field = e`) would need a more general solution, likely involving typing by constraints.


lib/OptimizeMiniRust.ml
```diff
@@ -1093,6 +1118,16 @@ let infer_mut_borrows files =
...
+        (* We need to perform the inference twice to handle function parameters
+           that are becoming mutable borrows, and that are assigned to in the function.
+
+           The first execution will update the function type to mark the parameter as
+           a mutable variable of type mutable borrow.
+           The second execution will use this information when building the initial
+           `known` set, forcing expressions assigned to the mutable parameter to
+           be mutable borrows.
+        *)
+        let decl = infer_function env valuation decl in
```
msprotz, line 1130: based on your comment it looks like what you want is a first List.map followed by a second one

msprotz (commented):
This looks good but breaks HACL -- let's see what the best fix is

lib/OptimizeMiniRust.ml
R1kM: Correct, this is now fixed

R1kM:
With the latest changes, HACL-rs compiles and verifies tests again. Required changes were:
* Handle the `split_at_mut` methodcall. As we now perform inference twice, several `split_at` were translated into `split_at_mut`.
* Refine the handling of pattern-matching: we populate the `known` map with field patterns that are already marked as ref or mut to avoid reinserting a dereference on field occurences.

lib/OptimizeMiniRust.ml
```diff
@@ -578,6 +601,7 @@ let rec infer_expr (env: env) valuation (return_expected: typ) (expected: typ) (
                 match pat with
                 | OpenP open_var ->
                     let { atom; _ } = open_var in
+                    let is_ref = VarSet.mem atom known.p in
```
msprotz, line 604: nit: rename to `was_ref`

msprotz (approved):
Approved, but i) with naming suggestion and ii) handling from_mut since I'm pretty sure this is going to break my WIP with SymCrypt... you probably need something like

```diff
diff --git a/lib/OptimizeMiniRust.ml b/lib/OptimizeMiniRust.ml
index dfc0b42a..cc2eabf0 100644
--- a/lib/OptimizeMiniRust.ml
+++ b/lib/OptimizeMiniRust.ml
@@ -285,6 +285,11 @@ let rec infer_expr (env: env) valuation (return_expected: typ) (expected: typ) (
         else
           let known, e1 = infer_expr env valuation return_expected (Ref (None, Shared, t)) known e1 in
           known, Call (Name [ "std"; "slice"; "from_slice" ], [ t ], [ e1 ])
+      else if n = [ "std"; "slice"; "from_mut" ] then
+        let e1 = KList.one es in
+        let t = KList.one targs in
+        let known, e1 = infer_expr env valuation return_expected (Ref (None, Shared, t)) known e1 in
+        known, Call (Name [ "std"; "slice"; "from_slice" ], [ t ], [ e1 ])
       else
         (* TODO: substitute targs in ts -- for now, we assume we don't have a type-polymorphic
            function that gets instantiated with a reference type *)
```

once this is fixed, I'll confirm this doesn't break symcrypt before hitting the merge button -- thanks!

lib/OptimizeMiniRust.ml
R1kM: Done in [581e262](https://github.com/FStarLang/karamel/pull/572/commits/581e262babea8fbcf64e1f3cb758f89ee2839e3d)

R1kM:
I pushed your suggested patch.
I am however a bit confused by the handling of `slice::from_mut`. If I am understanding correctly, a `from_mut` will now be translated to a `from_slice`, thus losing mutability. Shouldn't the type passed by a `Ref (None, Mut)`, and the returned expression be a `from_mut` Call?

msprotz:
Sorry this is a bad copy-paste I copied the wrong two lines above

```diff
diff --git a/lib/OptimizeMiniRust.ml b/lib/OptimizeMiniRust.ml
index dfc0b42a..a95e46c3 100644
--- a/lib/OptimizeMiniRust.ml
+++ b/lib/OptimizeMiniRust.ml
@@ -285,6 +285,11 @@ let rec infer_expr (env: env) valuation (return_expected: typ) (expected: typ) (
         else
           let known, e1 = infer_expr env valuation return_expected (Ref (None, Shared, t)) known e1 in
           known, Call (Name [ "std"; "slice"; "from_slice" ], [ t ], [ e1 ])
+      else if n = [ "std"; "slice"; "from_mut" ] then
+        let e1 = KList.one es in
+        let t = KList.one targs in
+        let known, e1 = infer_expr env valuation return_expected (Ref (None, Mut, t)) known e1 in
+        known, Call (Name [ "std"; "slice"; "from_mut" ], [ t ], [ e1 ])
       else
         (* TODO: substitute targs in ts -- for now, we assume we don't have a type-polymorphic
            function that gets instantiated with a reference type *)
            ```

PR #574, opened 2025-04-11
Add missing recursive calls when anonymizing unions

mtzguido:
We ran into this bug today (see testcase). I think this is the fix, the one thing I'm not sure about is whether passing `t2.typ` is correct. It seems like it's meant to be the type of the expression being visited, but I don't understand OCaml visitors well enough to be sure.


msprotz:
Yeah but there's a helper called visit_expr_w that does this for you automatically (while also calling the visitor on the type itself). Other than that looks great I remember hitting this bug years ago and working around it by not using -fanonymous-unions

mtzguido:
Nice, thanks! Updated.

msprotz (approved):

PR #577, opened 2025-04-15
Handle nested occurrences of pointers with `-fkeep-tuples`

gebner:
Without this change, pointers inside tuples would not cause lifetime to be emitted, for example:
```rust
  struct Foo {
    bar: (u8, &[u8]),
  }
```


msprotz (approved):

msprotz:
@gebner can I let you resolve conflicts and hit the merge button once you're ready? thanks

PR #582, opened 2025-04-22
Support for local function calls in MiniRust AST -- this fixes #581

### LINKED ISSUE #581: Rust: support function pointers
The C backend already supports function pointers to global functions.  EverCDDL makes use of that feature to implement array and map iterators.

It would be great if we could support function pointers (`fn(Foo) -> Bar`) in the Rust backend as well.

msprotz:
See the comments in the diff -- this unblocks you @gebner but if you have a function that takes a mutable reference then the mutability of the borrow will be inferred incorrectly.


msprotz:
CC @R1kM who might be interested in what could possibly be done here for the mutability analysis

PR #585, opened 2025-04-23
Infer Copy trait for functions.

gebner:


lib/OptimizeMiniRust.ml
```diff
@@ -1369,6 +1369,8 @@ let compute_derives files =
             self#plus (self#visit_typ () t) (if m = Mut then TraitSet.singleton PartialEq else everything)
           method! visit_Vec _ t =
             self#plus (self#visit_typ () t) (TraitSet.of_list [ PartialEq; Clone ])
+          method! visit_Function _ _ _ _ =
+            everything
```
msprotz, line 1373: functions have PartialEq and Clone too?

gebner: The test passes, so yes.

Using `==` on functions produces a warning:
> warning: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique

And Copy implies Clone.

PR #589, opened 2025-04-23
Apply -fcontained-type transformation in function types

### LINKED ISSUE #587: Rust: more fine-grained lifetimes in function pointers
To get the function pointers in EverCDDL to compile, we'd need slightly more sophisticated lifetimes in function pointers.

What Karamel produces out-of-the-box is the following:
```rust
    pub cddl_array_iterator_impl_validate:
    fn (&'a [crate::cbordetveraux::cbor_raw_iterator__CBOR_Pulse_Raw_Type_cbor_raw <'a>]) -> bool,
```

However, the slice must have a different (shorter) lifetime than the CBOR objects inside.  So we'd need the following, written using explicit lifetimes:
```rust
    pub cddl_array_iterator_impl_validate:
    for<'b> fn (&'b mut [crate::cbordetveraux::cbor_raw_iterator__CBOR_Pulse_Raw_Type_cbor_raw<'a>]) -> bool,
```

It would also be ok in this case to just drop the lifetimes and let Rust infer them:
```rust
    pub cddl_array_iterator_impl_validate:
    fn (&mut [crate::cbordetveraux::cbor_raw_iterator__CBOR_Pulse_Raw_Type_cbor_raw]) -> bool,
```

gebner:
The flag `-fcontained-type mytype` instructs the Rust backend to use a fresh lifetime name for the slice in function arguments like `f : slice mytype -> unit`.

This PR applies the same logic to function pointer types as well.

Fixes #587

### DIFF
--- a/lib/AstToMiniRust.ml
+++ b/lib/AstToMiniRust.ml
@@ -458,6 +458,10 @@ let default_config = {
   keep_mut = false;
 }
 
+let is_contained t =
+  let t = KList.last t in
+  List.exists (fun t' -> KString.starts_with t t') !Options.contained
+
 let rec translate_type_with_config (env: env) (config: config) (t: Ast.typ): MiniRust.typ =
   match t with
   | TInt w -> Constant w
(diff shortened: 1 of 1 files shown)


lib/AstToMiniRust.ml
```diff
@@ -488,7 +492,17 @@ let rec translate_type_with_config (env: env) (config: config) (t: Ast.typ): Min
   | TArrow _ ->
       let t, ts = Helpers.flatten_arrow t in
       let ts = match ts with [ TUnit ] -> [] | _ -> ts in
-      Function (0, List.map (translate_type_with_config env config) ts, translate_type_with_config env config t)
+      let translate_parameter_type t =
+        match translate_type_with_config env config t with
+        | Ref (_, m, (Slice (Name (t, _ :: _)) as slice_t)) when is_contained t ->
+            (* Unlike top-level declarations (where we use 'b), we omit the lifetime name here.
+               Writing an explicit lifetime name here would also require adding a for<'b>.
+               By omitting the lifetime name, Rust will automatically do this. *)
```
gebner, line 500: FWIW, we could also omit the explicit `'b` lifetime for top-level declarations as well.

msprotz:
Looks good thank you

PR #593, opened 2025-04-28
Specify necessary lifetimes in function definitions.

### LINKED ISSUE #588: Rust: no-box and keep-tuples introduce ambiguous lifetimes
Extracting the following file with `-fkeep-tuples` and `-fno-box` results in an error from the Rust compiler:

```fstar
module Rustamblifetime
type t = LowStar.Buffer.buffer bool
let foo (x: t) (y: t) : t & t = x, y
let main_ () = 0ul
```

The extracted Rust code needs to specify explicit lifetimes, but doesn't:
```rust
pub fn foo(x: &[bool], y: &[bool]) -> (&[bool], &[bool]) { (x,y) }
```

Rust raises the following error:
```
error[E0106]: missing lifetime specifiers
  --> .output/rustamblifetime.rs:11:40
   |
11 | pub fn foo(x: &[bool], y: &[bool]) -> (&[bool], &[bool]) { (x,y) }
   |               -------     -------      ^        ^ expected named lifetime parameter
   |                                        |
   |                                        expected named lifetime parameter
   |
   = help: this function's return type contains a borrowed value, but the signature does not say whether it is borrowed from `x` or `y`
help: consider introducing a named lifetime parameter
   |
11 | pub fn foo<'a>(x: &'a [bool], y: &'a [bool]) -> (&'a [bool], &'a [bool]) { (x,y) }
   |           ++++     ++             ++              ++          ++

error: aborting due to 1 previous error
```

gebner:
Function definitions now use explicit lifetime parameters if and only if the return type contains a pointer (potentially nested in a tuple).

This PR slightly simplifies the existing behavior, which used explicit lifetimes if one of the arguments was a named type with a lifetime (but not for slices).  Now it always omits the lifetimes if pointers only appear in arguments (but not in the return type):
```rust
pub fn add_carry_u32(x: u32, y: u32, r: &mut [u32], p: &[u32]) -> u32
```

Fixes #588.

### DIFF
--- a/lib/DataTypes.ml
+++ b/lib/DataTypes.ml
@@ -550,12 +550,12 @@ let compile_simple_matches ((map, enums, tag_remap): map) = object(self)
       let typ = self#visit_typ env x.typ in
       { node; typ; meta = x.meta }
 
-  method! visit_TQualified _ lid =
+  method! visit_TQualified env lid =
     match Hashtbl.find map lid with
     | exception Not_found ->
         TQualified lid
     | Eliminate t ->
-        t
+        self#visit_typ env t
     | _ ->
         TQualified lid
 
--- a/test/Rustabbrev2.fst
+++ b/test/Rustabbrev2.fst
@@ -1,0 +1,10 @@
+module Rustabbrev2
+
+type foo = { b: bool; }
+type bar = { c: foo; }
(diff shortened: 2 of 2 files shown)


msprotz (approved):
Looks good to me -- change makes sense

PR #595, opened 2025-04-29
Infer mutability in function pointer fields.

### LINKED ISSUE #590: Rust: mutability analysis for function pointer types
This was left out of #582.

I don't know enough Low* to write a test, but the issue in EverCDDL is the following.  Karamel extracts the following pointer type (after #589):
```rust
#[derive(PartialEq, Clone, Copy)]
pub struct arr_iter<'a> {
    pub cddl_array_iterator_impl_validate:
    fn (&[crate::cbordetveraux::cbor_raw_iterator__CBOR_Pulse_Raw_Type_cbor_raw <'a>]) -> bool,
    // ...
}
```

However, when we construct a value of this type, the validate function needs a mutable slice:
```rust
pub fn aux_env5_validate_1(
    pi: &mut [crate::cbordetveraux::cbor_raw_iterator__CBOR_Pulse_Raw_Type_cbor_raw]
) ->
    bool

// ...

        arr_iter
        {
            cddl_array_iterator_impl_validate: aux_env5_validate_1,
        };
```

We also have call sites to this function like this:
```rust
    let pj: [crate::cbordetveraux::cbor_raw_iterator__CBOR_Pulse_Raw_Type_cbor_raw; 1] =
        [i.cddl_array_iterator_contents; 1usize];
    let _test: bool = (i.cddl_array_iterator_impl_validate)(&pj);
```

The generated code compiles if I make two changes:
 1. Change the `&` to `&mut` in the pointer type.
 2. Use `let mut` and `&mut` at the call site.

gebner:
This implements a MVP for mutability inference in function pointer fields.  It's enough to extract EverCOSE to Rust.  But there's some limitations:
 - This is only for function pointers in structures; function pointers passed as arguments etc. are not supported yet.
 - Rust does not have variance for mutability, so if you want to assign a function `fn foo(a: &[u32]) -> bool` to a field with type `fn(&mut [u32]) -> bool`, then you need to mark the argument `a` of `foo` as mutable.  This is not implemented yet.

Fixes #590.  See also preceding PR #582.

cc @R1kM

### DIFF
--- a/lib/OptimizeMiniRust.ml
+++ b/lib/OptimizeMiniRust.ml
@@ -172,6 +172,22 @@ let add_mut_field (n: DataType.t) f =
     if sf.name = f then {sf with typ = fst (make_mut_borrow sf.typ) } else sf) fields in
   Hashtbl.replace structs n fields
 
+(* Update mutability of the arguments in a function type. *)
+let update_mut_args_fn_typ (t: typ) new_arg_tys : typ =
+  match t with
+  | Function (i, old_arg_tys, old_ret_ty) ->
+    Function (i, List.map2 (fun old_arg new_arg ->
+      if is_mut_borrow new_arg then fst (make_mut_borrow old_arg) else old_arg)
+      old_arg_tys new_arg_tys, old_ret_ty)
+  | _ -> t
+
+(* Update mutability of the arguments in a function pointer field. *)
+let update_mut_args_fn_field (n: DataType.t) fn new_arg_tys =
+  let fields = Hashtbl.find structs n in
+  let fields = List.map (fun (sf: MiniRust.struct_field) ->
+    if sf.name = fn then {sf with typ = update_mut_args_fn_typ sf.typ new_arg_tys } else sf) fields in
+  Hashtbl.replace structs n fields
+
 let retrieve_pair_type (t: typ) =
   match t with
   | Tuple [e1; e2] -> assert (e1 = e2); e1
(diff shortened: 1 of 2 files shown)


lib/OptimizeMiniRust.ml
```diff
@@ -315,6 +331,21 @@ let rec infer_expr (env: env) valuation (return_expected: typ) (expected: typ) (
...
+  | Call (Field (_, f, st) as fld, [], args) ->
+    let fields = Hashtbl.find structs (`Struct (assert_name st)) in
+    let field = List.find (fun fld -> fld.name = f) fields in
+    begin match field.typ with
+    | Function (_, arg_tys, _) ->
+      let known, args = List.fold_left2 (fun (known, args) arg arg_ty ->
+          let known, arg = infer_expr env valuation return_expected arg_ty known arg in
+          known, arg :: args
+        ) (known, []) args arg_tys
+      in
```
msprotz, line 343: aren't you missing a `List.rev args` here?

msprotz (approved):
This is a net improvement, so let's merge this, but please check my remark about List.rev. Thank you!

lib/OptimizeMiniRust.ml
gebner: You're totally right, that's why I usually stay far away from folds...

msprotz: What do you typically use then?

gebner: In cases like this, map + mutable reference/state monad.

Using combinators like map and filters has the obvious advantage that you can't screw up the order or forget elements, and you don't need to do the accumulator yourself.  Reasoning about filter/map is easier too, since it's just first order and you don't need to figure out loop invariants.

PR #606, opened 2025-05-20
Add missing case to Rust mutability analysis

R1kM:
Handle the `*atom = e` case in the mutability inference analysis.


msprotz (approved):
Can't believe this never popped up before! Thanks.

Also a good reminder to try to make this code more generic

R1kM:
I had pretty much the same reaction on both counts

PR #613, opened 2025-06-13
added support for float & double

ssyram:
As the title, support for float & double, namely, `Float32` and `Float64` are added to the `Constant.ml`. Relevant parts are fixed. This is required from the project [Eurydice](https://github.com/AeneasVerif/eurydice).

Question left: where to add the `float32_t` and `float_64_t` definitions? Just adding to Eurydice is fine, but if this is not defined, Krml might be less self-contained.

### DIFF
--- a/lib/Constant.ml
+++ b/lib/Constant.ml
@@ -8,10 +8,10 @@ type t = width * string
 
 and width =
   | UInt8 | UInt16 | UInt32 | UInt64 | Int8 | Int16 | Int32 | Int64
-  | Float32 | Float64 (** Floating point numbers. *)
   | Bool
   | CInt (** Checked signed integers. *)
   | SizeT | PtrdiffT
+  | Float32 | Float64 (** Floating point numbers. *)
 
 let bytes_of_width (w: width) =
   match w with
--- a/lib/Constant.ml
+++ b/lib/Constant.ml
@@ -66,9 +66,9 @@ let unsigned_of_signed = function
   | Float32 | Float64 -> raise (Invalid_argument "unsigned_of_signed")
 
 let is_signed = function
-  | Int8 | Int16 | Int32 | Int64 | CInt | PtrdiffT | Float32 | Float64 -> true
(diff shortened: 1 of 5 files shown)


lib/Constant.ml
```diff
@@ -61,10 +62,11 @@ let unsigned_of_signed = function
   | Int16 -> UInt16
   | Int32 -> UInt32
   | Int64 -> UInt64
-  | CInt | UInt8 | UInt16 | UInt32 | UInt64 | SizeT | PtrdiffT | Bool -> raise (Invalid_argument "unsigned_of_signed")
+  | CInt | UInt8 | UInt16 | UInt32 | UInt64 | SizeT | PtrdiffT | Bool
+  | Float32 | Float64 -> raise (Invalid_argument "unsigned_of_signed")
 
 let is_signed = function
-  | Int8 | Int16 | Int32 | Int64 | CInt | PtrdiffT -> true
+  | Int8 | Int16 | Int32 | Int64 | CInt | PtrdiffT | Float32 | Float64 -> true
```
msprotz, line 69: should go in the invalid_argument case for consistency with unsigned_of_signed, above

msprotz (commented):
Please fix my comment -- the definitions for float32 and float64 should go in `types.h` please. Thank you!

lib/Constant.ml
ssyram: Thanks, resolved.

lib/PrintCommon.ml
```diff
@@ -33,6 +33,8 @@ let width_to_string = function
   | Bool -> "bool"
   | SizeT -> if !Options.linux_ints then "size_t" else "size"
   | PtrdiffT -> if !Options.linux_ints then "ptrdiff_t" else "ptrdiff"
+  | Float32 -> "float32"
```
msprotz, line 36: you could use `float` and `double` here since they are guaranteed to be 32 and 64 bits respectively

ssyram:
Fixed.

PR #616, opened 2025-07-07
Introduce -faggressive-inlining

mtzguido:
Hi Jonathan, this is a revival or PR https://github.com/FStarLang/karamel/pull/469. We found that this flag can indeed be useful to reuse a single karamel file for both C and Rust. We want to turn this inlining feature on in C, but not in Rust as it breaks some code due to affecting lifetimes.

--

This flag makes karamel attempt to inline every intermediate letbinding, instead of only those marked with `@@CInline` or with some particular name.

There is an F* flag that is similar (`--ext extraction_inline_all`) that adds a `@@CInline` attribute to every letbinding, but using that flag means the choice of whether to inline or not makes it into the .checked file, and changing it requires re-typechecking. Further, it means we cannot reuse a single krml file to obtained both inlined and non-inlined code, for example C and Rust.

### DIFF
--- a/test/AggressiveInline.fst
+++ b/test/AggressiveInline.fst
@@ -1,0 +1,16 @@
+module AggressiveInline
+
+(* None of {foobar,temp_x,temp_y} should appear in the output. This is similar
+to the test in InlineLet.fst, but is extracted with the `-faggressive-inling`
+flag of karamel. So there is no need for the CInline attributes. *)
+
+let main () =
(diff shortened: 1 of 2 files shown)


msprotz (approved):
Fine by me -- there are no tests, so this might regress. If you have good coverage for this option elsewhere, then I'm happy. Thanks!

mtzguido:
You're right, thanks, added a small test.

PR #618, opened 2025-07-11
Fixed Eurydice error on floating point

ssyram:
During the C* -> C11 translation, `is_signed` will be called on floating point, and we should not raise error for that. This PR fixed that.

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -67,6 +67,8 @@ and type_scheme = {
 (* The visitor of types composes with the misc. visitor. *)
 and typ =
   | TInt of width
+      (** this constructor includes also **floating points** for historical
+       * reason. *)
   | TBool
   | TUnit
   | TAny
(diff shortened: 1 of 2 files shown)


msprotz:
I'd rather fix the caller please -- how come someone is asking whether Float is signed before checking that it makes sense to do so? Can you give a backtrace for the issue?

ssyram:
I don't know... But the original Eurydice code complained about this, it says `is_signed` is called in Krml.

ssyram:
Shall we change `TInt` into `TIntOrFloat`? Or can we simply add `TFloat` to denote this?

msprotz:
I think it's ok for now. I only see a one-line change in this PR -- can you add the patch that we talked about on zulip?

ssyram:
I think the only place here is TInt, which uses the name Int to include also Float... So just one hint is fine? Constant is a more general name and width is also fine.

msprotz (approved):
(I ended up merging my suggested patch on another PR -- sorry for the confusion)

ssyram:
OK, thanks! Shall we merge in Eurydice now?

PR #623, opened 2025-07-13
Make sure to not duplicate non-constant VLA lengths

### LINKED ISSUE #621: Length of local arrays duplicated in generated code
For this F* code
```fstar
module Malloca

open FStar.HyperStack
open FStar.HyperStack.ST

inline_for_extraction noextract
let gen (f : unit -> UInt32.t) : Stack Int32.t (fun _ -> False) (fun h0 _ h1 -> True) =
  let b = FStar.Buffer.create 1uy (f ()) in
  42l

assume val f () : UInt32.t

let test () : Stack Int32.t (fun _ -> False) (fun _ _ _ -> True) =
  gen f
```
we generate the following C
```c
int32_t Malloca_test(void)
{
  KRML_CHECK_SIZE(sizeof (uint8_t), Malloca_f());
  uint8_t b[Malloca_f()];
  memset(b, 1U, Malloca_f() * sizeof (uint8_t));
  KRML_MAYBE_UNUSED_VAR(b);
  return (int32_t)42;
}
```
Note the repeated calls to `Malloc_f()`. I was about to write that this may be correct, only bad for performance, but this seems to happen even when the length is effectful:
```fstar
inline_for_extraction noextract
let gen (f : unit -> Stack UInt32.t (fun _ -> False) (fun _ _ _ -> True)) : Stack Int32.t (fun _ -> False) (fun h0 _ h1 -> True) =
  let b = FStar.Buffer.create 1uy (f ()) in
  42l

let f () : Stack UInt32.t (fun _ -> False) (fun _ _ _ -> True) = 1ul

let test () : Stack Int32.t (fun _ -> False) (fun _ _ _ -> True) =
  gen f
```
gives:
```c

uint32_t Malloca_f(void)
{
  return 1U;
}

int32_t Malloca_test(void)
{
  KRML_CHECK_SIZE(sizeof (uint8_t), Malloca_f());
  uint8_t b[Malloca_f()];
  memset(b, 1U, Malloca_f() * sizeof (uint8_t));
  KRML_MAYBE_UNUSED_VAR(b);
  return (int32_t)42;
}
```
My Low* is too rusty to write something that will actually make this crash, but I'm sure one can write an `f` that returns different values. Regardless, this can be broken when `f` is just a declaration for an external function.

This is what the karamel file looks like:
```
$ ./bin/fstar.exe --read_krml_file Malloca.krml 
Karamel format version: 31
Malloca:
(DFunction None [] 0 (TInt UInt32) (["Malloca"], "f")
  [{ name = "uu___"; typ = TUnit; mut = false; meta = [] }]
  (EConstant (UInt32, "1")))

(DFunction None [] 0 (TInt Int32) (["Malloca"], "test")
  [{ name = "uu___"; typ = TUnit; mut = false; meta = [] }]
  (ELet { name = "b"; typ = (TBuf (TInt UInt8)); mut = false; meta = [] }
    (ELet { name = "uu___1"; typ = (TInt UInt32); mut = false; meta = [] }
      (EApp (EQualified (["Malloca"], "f")) [EUnit])
      (EBufCreate Stack (EConstant (UInt8, "1")) (EBound 0)))
    (EConstant (Int32, "42"))))
```

(I'm not using Low*, I noticed this from Pulse, but figured this is a simpler test case.)

mtzguido:
Probably deserves a review, not sure if style is right or if there is a better palce for this.

Fixes #621

### DIFF
--- a/lib/CStarToC11.ml
+++ b/lib/CStarToC11.ml
@@ -535,18 +535,24 @@ and ensure_array_l t inits =
   | _ ->
       t
 
-and ensure_array m t expr =
+(* Returns
+   - C* array type
+   - C* initializer
+   - C* num elems
+   - C11 (translated) size of elem
+ *)
+and ensure_array m (t : CStar.typ) (expr : CStar.expr) : CStar.typ * CStar.expr * CStar.expr * C.expr =
   let mul x y = C.Op2 (K.Mult, x, y) in
   match t, expr with
-  | Pointer t, BufCreate ((Stack | Eternal), init, size)  ->
-      let t, init, size' = ensure_array m t init in
-      Array (t, size), init, mul (mk_expr m size) size'
-  | Array (t, l), BufCreate ((Stack | Eternal), init, size) ->
-      assert (l = size);
-      let t, init, size' = ensure_array m t init in
-      Array (t, size), init, mul (mk_expr m size) size'
+  | Pointer t, BufCreate ((Stack | Eternal), init, len)  ->
+      let _, init, len', size' = ensure_array m t init in
+      Array (t, len), init, len, mul (mk_expr m len') size'
+  | Array (t, l), BufCreate ((Stack | Eternal), init, len) ->
+      assert (l = len);
+      let _, init, len', size' = ensure_array m t init in
+      Array (t, len), init, len, mul (mk_expr m len') size'
   | _ ->
-      t, expr, C.Sizeof (C.Type (mk_type m t))
+      t, expr, CStar.Constant (K.UInt8, "1"), C.Sizeof (C.Type (mk_type m t))
 
 and decay_array t =
   match t with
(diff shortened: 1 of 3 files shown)


lib/CStarToC11.ml
```diff
@@ -641,16 +647,30 @@ and mk_stmt m (stmt: stmt): C.stmt list =
...
        * declare a fixed-length array; this is an "upcast" from pointer type to
        * array type, in the C sense. *)
-      let t, init, size = ensure_array m binder.typ rhs in
-      (* TODO: make ensure_array return both *)
-      let n_elements = match size with
-        | C.Op2 (K.Mult, outer_size, _) -> outer_size
-        | _ -> failwith "impossible, see ensure_array"
+      let t, init, n_elements, e_size = ensure_array m binder.typ rhs in
+
+      let is_constant = match n_elements with Constant _ -> true | _ -> false in
```
msprotz, line 652: you could have a slightly better criterion here, and include variables, or arithmetic expressions -- this is going to generate a diff in HACL's bignums otherwise (which use VLAs in a limited number of places)

msprotz (approved):
Looks good apart from my minor comment. Thank you!

lib/CStarToC11.ml
msprotz: yeah but then you'd have to go in depth to make sure that no `Call` nodes are hidden in there... might be a little bit of engineering -- I'm fine either with a simple criterion (`let rec is_simple_expression`) or a more general one (`object inherit [_] reduce_cstar method! visit_BufCreate _ _ _ _ = false method! visit_BufCreateL _ _ _ _ = false method! visit_Call _ _ _ = false etc.`)

mtzguido: Pushed an update (without objects...)

msprotz (approved):

PR #629, opened 2025-07-18
Introduce -dbt: print a backtrace on an error

mtzguido:
This allows to pass -dbt to get a stack trace on an error:
```
$ krml -dbt [..] foo.krml
[...]
This is not a buffer: any
Warning 4 is fatal, exiting.
Stack trace:
Raised at Krml__Checker.checker_error.(fun) in file "lib/Checker.ml", line 85, characters 4-104
Called from Krml__Checker.infer' in file "lib/Checker.ml", line 700, characters 18-37
Called from Krml__Checker.infer in file "lib/Checker.ml", line 527, characters 10-22
Called from Stdlib__List.map in file "list.ml", line 86, characters 15-19
Called from Krml__Checker.infer' in file "lib/Checker.ml", line 776, characters 13-38
Called from Krml__Checker.infer in file "lib/Checker.ml", line 527, characters 10-22
Called from Krml__Checker.infer' in file "lib/Checker.ml", line 638, characters 13-26
Called from Krml__Checker.infer in file "lib/Checker.ml", line 527, characters 10-22
Called from Stdlib__List.map in file "list.ml", line 83, characters 15-19
Called from Krml__Checker.infer' in file "lib/Checker.ml", line 649, characters 16-39
Called from Krml__Checker.infer in file "lib/Checker.ml", line 527, characters 10-22
Called from Krml__Checker.check' in file "lib/Checker.ml", line 335, characters 8-21
Called from Krml__Checker.check in file "lib/Checker.ml", line 278, characters 2-16
Called from Krml__Checker.check_or_infer in file "lib/Checker.ml", line 228, characters 4-17
Called from Krml__Checker.check' in file "lib/Checker.ml", line 343, characters 15-80
Called from Krml__Checker.check in file "lib/Checker.ml", line 278, characters 2-16
Called from Krml__Checker.infer' in file "lib/Checker.ml", line 836, characters 6-23
Called from Krml__Checker.infer in file "lib/Checker.ml", line 527, characters 10-22
Called from Krml__Checker.infer' in file "lib/Checker.ml", line 647, characters 14-25
Called from Krml__Checker.infer in file "lib/Checker.ml", line 527, characters 10-22
Called from Krml__Checker.check' in file "lib/Checker.ml", line 335, characters 8-21
Called from Krml__Checker.check in file "lib/Checker.ml", line 278, characters 2-16
Called from Krml__Checker.check_program.(fun) in file "lib/Checker.ml", line 170, characters 6-22
```

For uncaught errors, setting `OCAMLRUNPARAM=b` already had the same effect (though one must remember this instead of being an actual option). This option makes it also easy to get stack traces for caught errors like warning 4 above.

### DIFF
--- a/lib/AstToCFlat.ml
+++ b/lib/AstToCFlat.ml
@@ -382,7 +382,7 @@ let populate env files =
           with e ->
             KPrint.beprintf "[AstToC♭] can't compute the layout of %a:\n%s\n%s"
               PrintAst.plid lid (Printexc.to_string e)
-              (if Options.debug "backtraces" then Printexc.get_backtrace () ^ "\n" else "");
+              (if !Options.backtrace then Printexc.get_backtrace () ^ "\n" else "");
             env
           end
       | _ ->
@@ -709,9 +709,8 @@ and mk_expr_or_bail (env: env) (locals: locals) (e: expr): locals * CF.expr =
         Hashtbl.replace errors !current_lid (s :: Hashtbl.find errors !current_lid)
       else
         Hashtbl.add errors !current_lid [ s ];
-      let bt = if Options.debug "backtraces" then Printexc.get_backtrace () ^ "\n" else "" in
-      let msg = KPrint.bsprintf "%a: compilation error turned to runtime failure\n%s%s"
-        plid !current_lid s bt
+      let msg = KPrint.bsprintf "%a: compilation error turned to runtime failure\n%s"
+        plid !current_lid s
       in
       mk_expr env locals (Helpers.with_unit (EAbort (None, Some msg)))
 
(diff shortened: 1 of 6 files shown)


msprotz (approved):
Thanks, should this be merged with the existing "-d backtraces", or the latter removed? (I think the latter is from a very long while ago where errors had to be recoverable because so much simply did not extract... might be no longer necessary)

mtzguido:
Ah I didn't know about that one! It seems to be printing backtraces when some exceptions are caught, which is good too. What about unifying all behavior into `-dbt`? (Or `-dbacktrace`, etc, let me know what name you'd prefer?)

There is one more use of `-d backtraces`: it is [adding some backtraces in generated code](https://github.com/FStarLang/karamel/blob/0bf25f35a58052b6148867d2449be5c2f82427f1/lib/AstToCFlat.ml#L712-L716). Maybe that can be removed and then it's clear that `-d bt` is debug-only?

msprotz:
- Yes to unifying the behavior -- I prefer -dbacktrace
- The wasm backend is on its way out so totally ok to get rid of this code
Thanks!

mtzguido:
Updated

msprotz (approved):

PR #632, opened 2025-07-24
Substitute cgs before types for all cases in Checker.ml

Lin23299:
In the following code : 
https://github.com/FStarLang/karamel/blob/d3f05e518a1d9aeee3acfa596c5bf205d6383cb0/lib/Checker.ml#L867-L879

I think we should substitute the cgs before the types. Because the types may include cgs comming from the current env, which maybe subtituted again by `subst_ctn'`. Thus cause a type mismatch fail. This actually happens when I testing the `[T;C] -> arr<T,C>` translation of arrays in Eurydice on the test case 'where_clauses_fncg.rs'.

I notice that here you actually runs `subst_ctn` before `subst_tn` with some debug infos:

https://github.com/FStarLang/karamel/blob/d3f05e518a1d9aeee3acfa596c5bf205d6383cb0/lib/Checker.ml#L588-L595

So I guess that maybe you've once triggered this bug and fixed it here. But the other cases in this file are not changed. This's why I open this PR for changing all other substitution orders instead of opening an issue.

### DIFF
--- a/lib/Checker.ml
+++ b/lib/Checker.ml
@@ -465,14 +465,14 @@ and check' env t e =
       | lid, ts, cgs when kind env lid = Some Record ->
           let fieldtyps = assert_flat env (lookup_type env lid) in
           let fieldtyps = List.map (fun (field, (typ, m)) ->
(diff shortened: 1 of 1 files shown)


msprotz:
Great, thank you!

PR #635, opened 2025-08-05
Option length for arrays in CStar and C11 for flexible array member

Lin23299:
Comes from the discussion in the PR  [Translating rust array type [T;C] into struct arr { T[C] data; } #252](https://github.com/AeneasVerif/eurydice/pull/252) for Eurydice. @msprotz


msprotz (approved):
Does this break Eurydice,

PR #641, opened 2025-08-12
Remove old Docker CI files and build_local scripts.

mtzguido:
(RFC, are the containers useful?)

The build_local scripts seem broken:
```
ERROR: failed to build: failed to solve: projecteverest/fstar-linux:94daa5edfc01: failed to resolve source metadata for docker.io/projecteverest/fstar-linux:94daa5edfc01: docker.io/projecteverest/fstar-linux:94daa5edfc01: not found
```
and I don't suspect they're very useful. Remove them, and all the other stale Dockerfiles (most of which were only for CI).

If a Dockerfile is useful, I can add one without this config.json business.


msprotz:
As far as I know no one really uses Docker for their dev workflow

PR #645, opened 2025-08-18
Checker: fix check in EBufCreate case

### LINKED ISSUE #644: Inlining alloc into free fails
In this snippet:
```fstar
let test1 () : Stack unit (fun _ -> False) (fun _ _ _ -> True) =
  let b = FStar.Buffer.create 1uL 999ul in
  FStar.Buffer.rfree b

let test2 () : Stack unit (fun _ -> False) (fun _ _ _ -> True) =
  [@@CInline] let b = FStar.Buffer.create 1uL 999ul in
  FStar.Buffer.rfree b
```
The second function fails to translate, saying:
```
Cannot re-check AllocFree.test2 as valid Low* and will not extract it.  If AllocFree.test2 is not meant to be extracted, consider marking it as Ghost, noextract, or using a bundle. If it is meant to be extracted, use -dast for further debugging.

Warning 4: in top-level declaration AllocFree.test2, in file AllocFree: Malformed input:

This subexpression creates a buffer with an unknown type:
  stack newbuf 1uint64_t 999uint32_t

Here's a hint. If you're using untagged unions, instead of:

  match e with
  | Foo ->
      Buffer.create e1 l1
  | Bar ->
      Buffer.create e2 l2

where e1: t1 and e2: t2, try:

  match e with
  | Foo ->
      let b: Buffer.buffer t1 = Buffer.create e1 l1 in
      b
  | Bar ->
      let b: Buffer.buffer t2 = Buffer.create e2 l2 in
      b

Warning 4 is fatal, exiting.
```

(This was triggered via Pulse and using -faggressive-inlining, but this example is smaller.) The krml input file looks fine:
```
$ fstar.exe --read_krml_file .output/AllocFree.krml 
Karamel format version: 31
AllocFree:
(DFunction None [] 0 TUnit (["AllocFree"], "test1")
  [{ name = "uu___"; typ = TUnit; mut = false; meta = [] }]
  (ELet { name = "b"; typ = (TBuf (TInt UInt64)); mut = false; meta = [] }
    (EBufCreate Stack (EConstant (UInt64, "1")) (EConstant (UInt32, "999")))
    (EBufFree (EBound 0))))

(DFunction None [] 0 TUnit (["AllocFree"], "test2")
  [{ name = "uu___"; typ = TUnit; mut = false; meta = [] }]
  (ELet
    { name = "b"; typ = (TBuf (TInt UInt64)); mut = false; meta = [CInline] }
    (EBufCreate Stack (EConstant (UInt64, "1")) (EConstant (UInt32, "999")))
    (EBufFree (EBound 0))))
```

I would say inlining this is fine, even if odd, but sometimes seems to not like it.

mtzguido:
It's the actual type of the buffer elements, not the "goal" type received by check, that must not be TAny.

Fixes #644

### DIFF
--- a/lib/Checker.ml
+++ b/lib/Checker.ml
@@ -369,7 +369,8 @@ and check' env t e =
         Warn.(maybe_fatal_error (loc, Vla e))
       end;
       let t, _ = assert_buffer env t in
-      if t = TAny then
+      let t' = check_or_infer env t e1 in
+      if t' = TAny then
         checker_error env buf_any_msg ppexpr e;
       check env t e1;
       check_array_index env e2;
--- a/test/AllocFree.fst
+++ b/test/AllocFree.fst
@@ -1,0 +1,14 @@
+module AllocFree
+
+open FStar.HyperStack
+open FStar.HyperStack.ST
+
+let test1 () : Stack unit (fun _ -> False) (fun _ _ _ -> True) =
+  let b = FStar.Buffer.rcreate_mm root 1uL 999ul in
+  FStar.Buffer.rfree b
+
+let test2 () : Stack unit (fun _ -> False) (fun _ _ _ -> True) =
+  [@@CInline] let b = FStar.Buffer.rcreate_mm root 1uL 999ul in
+  FStar.Buffer.rfree b
+
+let main () = 0l


msprotz:
Good catch, thanks! Surprised this never showed up before...

PR #656, opened 2025-11-06
Proper composition of the slice trick and monomorphized state resolut…

protz:
…ion -- fixes aeneasverif/eurydice#326


protz:
@microsoft-github-policy-service agree company="Google"

PR #661, opened 2025-11-16
opam MacOS support: Makefile: guard .SHELLSTATUS checks for MacOS

amosr:
Hi,

I had a few issues building karamel from opam on MacOS. One is because MacOS comes with a very old make. Using a new gnu make for development works fine, but it'd be convenient if opam could install without explicitly telling opam to use gmake.

Issue two is opam's sandboxing didn't like karamel's run-with-log command. I think it was writing to somewhere that opam didn't expect it to, but I'm not entirely sure. I just modified the .opam file to disable the logging while building via opam. I suspect this issue isn't limited to MacOS but not sure.

### DIFF
--- a/lib/AstToCStar.ml
+++ b/lib/AstToCStar.ml
@@ -375,7 +375,7 @@ and mk_expr env in_stmt under_initializer_list e =
   | EBound var ->
       CStar.Var (find env var)
   | EEnum lident ->
-      CStar.Qualified lident
+      CStar.Qualified (GlobalNames.mangle_enum lident e.typ)
   | EQualified lident ->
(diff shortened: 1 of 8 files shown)


amosr:
@microsoft-github-policy-service agree

Makefile
```diff
@@ -1,15 +1,20 @@
 # Including a Makefile from the visitors package, but making sure
 # to give decent errors.
 
+ifneq ($(shell uname),Darwin)
```
amosr: hmm, I guess I could clean this up by adding an option like `NO_SHELLSTATUS=1` to only skip this check in opam. open to suggestions

PR #662, opened 2025-11-18
Introduce a constant-folding / simplification phase

mtzguido:
This phase will fold operations between constants that we not reduced by F* (e.g. if they were enabled by inlining at the Karamel level) and also simplify expressions such as `x+0` or `x*1` which may appear at the C level.

The correctness of the constand folding relies on the absence of overflow and UB in these operations (except for the wrapping operators, these are handled well), and the fact the operations in C and OCaml's ZArith coincide. It's my understanding that the semantics of division and modulo in C and ZArith coincide
(https://ocaml.org/p/zarith/latest/doc/zarith/Z/index.html#val-div, C standard s6.5.6).

---

A skeptical eye would be good! Particularly wrt overflow, widths, etc.

If this isn't always desirable I could add an option.

I also have a patch to simplify `(a / n) * n + a % n ~> a`, but I thought it was a bit too specific. It's useful in my case (nvcc not simplifying it, making it much slower). Happy to add it here if that makes sense.

Here is the diff this would generate in HACL*: [diff.txt](https://github.com/user-attachments/files/23612152/diff.txt)

### DIFF
--- a/lib/Simplify.ml
+++ b/lib/Simplify.ml
@@ -415,6 +415,98 @@ let wrapping_arithmetic = object (self)
         EApp (self#visit_expr env e, List.map (self#visit_expr env) es)
 end
 
+let constant_fold = object (self)
+
+  inherit [_] map
+
+  method! visit_EApp env e es =
+    let fit (w:width) (x:Z.t) : Z.t =
+      (* Fit an integer into the allowed values for a width.
+         This only needs to handle widths for which there are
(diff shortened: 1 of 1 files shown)


protz:
This seems reasonable, because you only do this for homogeneous operations, meaning that we don't have to reason about promotion. Also, you only need to concern yourself with F*/krml semantics -- homogeneous operations on widths with wraparound modulo for unsigned variants, and matching that to C semantics is handled by AstToCStar.

PR #678, opened 2026-02-11
Fix Failure("nth") crash in remove_unused_parameters

### LINKED ISSUE #677: `remove_unused_parameters` crashes with `Failure("nth")` when EApp has more arguments than callee type arrow
## Summary

When bundling modules, `remove_unused_parameters` in `lib/Simplify.ml` crashes with `Failure("nth")` if a function call site has more actual arguments than the callee's formal type parameters (as reported by `flatten_arrow`). This happens when erased/ghost parameters are passed as `()` at call sites, but the callee's type signature doesn't include them (e.g., when the callee is an external or its type was simplified during extraction).

## Root Cause

In `lib/Simplify.ml`, the `unused` function (line 226) computes whether parameter `i` should be eliminated. At line 236:

```ocaml
List.nth ts i = TUnit
```

This accesses `ts` (the formal parameter type list) at index `i` without bounds checking. When called from `update_table_current` (line 344 in `visit_EApp`), the iteration is over the actual arguments `es`, so `i` can be `>= List.length ts`.

The length guard at line 353 (`if List.length es <= List.length ts then`) correctly handles this case for the main transformation, but `update_table_current` is called **before** this guard at line 344.

## Backtrace

```
Fatal error: exception Failure("nth")
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Krml__Simplify.unused.unused_i in file "lib/Simplify.ml", line 236
Called from Krml__Simplify.unused in file "lib/Simplify.ml", line 238
Called from Krml__Simplify.remove_unused_parameters.object#update_table_current.(fun) in file "lib/Simplify.ml", line 316
Called from Stdlib__List.iteri in file "list.ml", line 116
Called from Krml__Simplify.remove_unused_parameters.object#visit_EApp in file "lib/Simplify.ml", line 344
...
Called from Krml__Simplify.remove_unused_parameters.object#visit_DFunction in file "lib/Simplify.ml", line 256
```

## How to Reproduce

Extract a Pulse module with an `.fsti` interface (making implementations private in krml), create a wrapper module without `.fsti` that re-exports the functions, then bundle them together:

```bash
krml -skip-compilation -skip-makefiles -skip-linking \
  -bundle 'Wrapper=Wrapper,MyLib,MyLib.Spec' \
  -tmpdir output \
  Wrapper.krml MyLib.krml MyLib_Spec.krml
```

The crash triggers because after inlining, the bundled code contains calls to external functions (e.g., `Pulse.Lib.Array.PtsTo.op_Array_Access`) with 4 arguments (`arr, idx, (), ()`), but the callee's type only shows 2 parameters (the ghost `perm` and `seq` args were erased from the type but not from call sites).

## Proposed Fix

In `lib/Simplify.ml` line 236, add a bounds check:

```diff
-    List.nth ts i = TUnit
+    i < List.length ts && List.nth ts i = TUnit
```

This makes `unused_i` return `false` for out-of-bounds indices (i.e., "don't eliminate this parameter"), which is the correct conservative behavior.

I've verified this one-line fix resolves the crash and produces correct C output.

## Version

KaRaMeL commit: d6607b99477640cb1e5d423d5cbe709d76da61f7 (opam karamel 1.0.0)

elefthei:
Fixes #677

## Problem

`remove_unused_parameters` in `lib/Simplify.ml` crashes with `Failure("nth")` when bundling modules where a function call site has more actual arguments than the callee's formal type parameters (as reported by `flatten_arrow`).

This happens when erased/ghost parameters are passed as `()` at call sites, but the callee's type signature doesn't include them (e.g., external functions like `Pulse.Lib.Array.PtsTo.op_Array_Access` whose ghost args were erased from the type but not from call sites).

## Root Cause

In `unused_i` (line 236), `List.nth ts i = TUnit` is called without bounds checking. The `update_table_current` call at line 344 in `visit_EApp` iterates over actual arguments `es` and calls `unused arg_i` for each — but `arg_i` can exceed `List.length ts` when there are more actual args than formal type params. The length guard at line 353 correctly handles this for the main transformation path, but `update_table_current` is invoked *before* that guard.

## Fix

Add a bounds check so `unused_i` returns `false` for out-of-bounds indices (conservative: don't eliminate unknown params):

```diff
-    List.nth ts i = TUnit
+    i < List.length ts && List.nth ts i = TUnit
```

## Reproduction

Extract a Pulse module with `.fsti` interface (making implementations private), create a wrapper without `.fsti`, and bundle them:

```bash
krml -bundle 'Wrapper=Wrapper,Lib,Lib.Spec' Wrapper.krml Lib.krml Lib_Spec.krml
# Fatal error: exception Failure("nth")
```

After the fix, extraction succeeds and produces correct C output with all functions.


protz:
@tahina-pro is this something you can help with? I would love for this PR to have a corresponding regression in the test suite, which would allow me to understand the problem -- right now, without a supporting example, it's hard to determine whether this is the right fix or not

PR #679, opened 2026-02-11
Fix issue 676

protz:
I recommend viewing the diff with whitespace off: [link](https://github.com/FStarLang/karamel/pull/679/changes?w=1)

the changes are relatively minimal and consist in skipping a recursive traversal if we have no good reason to do it

this also allowed me to remove a call to self#visit_node that I always thought was un-necessary but for some reason remained... good riddance

@tahina-pro are there any additional regressions you would like to run for this?

on my end, this breaks eurydice: https://github.com/AeneasVerif/eurydice/pull/378 so I need to investigate what's going on there

### DIFF
--- a/lib/DataTypes.ml
+++ b/lib/DataTypes.ml
@@ -687,21 +687,13 @@ let remove_unit_buffers = object (self)
 
 end
 
-let remove_unit_fields = object (self)
+let build_unit_field_table erasable_fields = object (self)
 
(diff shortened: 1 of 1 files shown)


protz:
Things look good on the Eurydice end -- @tahina-pro please confirm this works for you in case you have other regression suites you'd like to run. Thank you!

PR #682, opened 2026-03-03
make: opam distribution on macs

amosr-msft:
Relax the opam build rules to always have a fallback option. Previously the constraints were too tight if on mac with OPAMMAKECMD set.

Disable the gtime check if NOSHORTLOG is set. Previously was failing on mac if gtime is not installed, but this is actually unused in opam with logging disabled.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,17 +18,9 @@ defaults:
     shell: bash
 
 jobs:
-  build-and-test:
+  build-deps:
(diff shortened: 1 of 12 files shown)


amosr-msft:
related pulse change: https://github.com/FStarLang/pulse/pull/582

protz:
It's not a super deep change and I work on Mac, so if you want to go ahead and merge this, do not hesitate

PR #683, opened 2026-03-08
Lef/lowstar dangling refs

elefthei:


elefthei:
@nikswamy looks like those were left behind from the fstar2 port

PR #701, opened 2026-03-27
Remove need for KRML_HOME

mtzguido:
Finally revived this idea. This changes krml to find the files it needs relatively to its binary location, instead of via KRML_HOME. Some notes about decisions I made along the way:

1- during building krmllib, we override this autodetection via some env vars, as karamel is just a dune-built executable in `_build/..` without its support files
2- added `-locate-include` and `-locate-krmllib` to use from Makefiles
3- the default build now produces a working karamel setup in `out/`
4- the top-level `krml` symlink still exists, with one caveat: after `make minimal` (and during the build of krmllib) it points to the dune executable. On a full build (after krmllib is build) it points to `out/lib/pulse`. I can see this being confusing, so I can change it if desired.
5- added a `make package` rule to create a tarball with a working karamel. This package looks the same as a local or OPAM install.
6- I moved `runtime` into `krmllib/runtime`. It was already installed that way (though I'm not sure anyone was installing karamel)
7- AutoConfig is gone.

I *think* client projects like EverParse will continue work if they set `KRML_HOME` and use `$KRML_HOME/krml`, since this binary will just do the right thing regardless of the variable. But I should confirm, hence marking as a draft.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,8 +39,9 @@ jobs:
         uses: actions/cache/restore@v4
         with:
           path: _opam
-          key: opam-${{ runner.os }}-${{ runner.arch }}-${{ github.run_id }}
+          key: opam-${{ runner.os }}-${{ runner.arch }}-fstar1-${{ github.run_id }}
           restore-keys: |
+            opam-${{ runner.os }}-${{ runner.arch }}-fstar1-
             opam-${{ runner.os }}-${{ runner.arch }}-
 
       # Build F*, from the fstar1 branch.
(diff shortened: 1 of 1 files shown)


protz (commented):
It seems like the tests now need a local install to succeed. That's not going to work for me, I don't think I should run `make local-install` to run the tests. Am I misunderstanding?

Generally, the idea behind the current solution is that it should not matter which executable you call krml through (_build/.../Karamel.exe or ./krml) -- either way should work. Is this property preserved?

I'll take a look into more detail next week. Thanks!

protz:
>  (I imagine yes.) 

I'm actually not sure. Most people who do repackaging end up writing their own Makefiles and/or changing the directory structure. Perhaps some sensible defaults would be fine, indeed.

protz:
- Reasonable
- eurydice and scylla both show some entries for `git grep AutoConfig`
- no worries about rebasing

PR #702, opened 2026-03-27
CI: Test Pulse with fstar2

mtzguido:
This adds a separate job to build fstar2 + test some Pulse examples (in test/pulse). These are not run by the normal make test to not break the current status of being able to build + test with fstar1. The new job is required to pass CI and be able to merge.

The new job tries to reuse the OPAM cache from the existing build-deps, but writing it back to into a separate key. 

The tests themselves are taken from Pulse's existing unit tests, and this brings in a copy of test.mk to drive things.

Here's a run with caches full: https://github.com/mtzguido/karamel/actions/runs/23667959139

### DIFF
--- a/test/pulse/Break.expected.c
+++ b/test/pulse/Break.expected.c
@@ -42,17 +42,17 @@ void Break_break_continue_and_return(uint8_t which)
   while (cond)
   {
     bool _continue = false;
-    if (which == 0U)
+    if ((uint32_t)which == 0U)
       _break = true;
     bool _break1 = _break;
     if (!_break1)
     {
-      if (which == 1U)
+      if ((uint32_t)which == 1U)
         _continue = true;
       bool _continue1 = _continue;
       if (!_continue1)
       {
-        if (which == 2U)
+        if ((uint32_t)which == 2U)
           _return = true;
         bool _return1 = _return;
         if (!_return1)


protz:
> But the dependencies of this file are huge (157 in total)

That's what I do for the regular test suite: I do a regular `fstar --dep` call, then extract all of the dependencies. Because it's a very parallel job, it doesn't take that long to build. So, up to you, since ultimately you'll have to maintain this file.

I'm not too adamant about the last point about recursive makes in subdirectories, except that if you end up deciding to share the extraction of the krml files, then you'd benefit from not having independent makefiles in subdirectories (so that the build of each test knows about the construction of its dependencies).

protz:
But just to clarify my understanding: you are *not* relying on recursive make here, it's just a single Makefile for the test suite?

mtzguido:
> That's what I do for the regular test suite: I do a regular `fstar --dep` call, then extract all of the dependencies. Because it's a very parallel job, it doesn't take that long to build. So, up to you, since ultimately you'll have to maintain this file.

Yeah, this sounds better. I think I'm hitting a bug in F* related to the removal of --cmi preventing me from doing this (it complains about polymorphic assumes on functions that are in fact implemented), but I will try to make it work soon.

Do you think we could merge meanwhile? That way we can advance F* and not break CI here.

> I'm not too adamant about the last point about recursive makes in subdirectories,

That's fair

> But just to clarify my understanding: you are not relying on recursive make here, it's just a single Makefile for the test suite?

Correct, all recursion is gone, I was just describing F*.

Issue #693, opened 2026-03-23
[DeepTest] Chained bitwise XOR/OR after wrapping arithmetic produces wrong results


elefthei:
## Bug Description

When UInt8/UInt16 wrapping arithmetic (`add_mod`, `sub_mod`, `mul_mod`) feeds into bitwise XOR (`^^`) or OR (`|^`), the uint32 intermediate carries extra upper bits that corrupt the bitwise result.

`BOr` and `BXor` are in the "no mask" category of `mk_arith` (`lib/AstToCStar.ml:339`), so the arithmetic result is not masked before the bitwise operation.

Note: Bitwise AND (`&^`) is **NOT** affected because AND naturally masks upper bits.

## Reproduction

**F\* code** (verifies successfully):
```fstar
open FStar.UInt8

let add_xor_compare (a b c d: UInt8.t) : Tot bool =
  ((a +%^ b) ^^ c) = d

// F*: (200+100) mod 256 = 44, 44 XOR 0xFF = 211
let lemma_add_xor () : Lemma (add_xor_compare 200uy 100uy 0xFFuy 211uy = true) = admit ()

let add_or_compare (a b c d: UInt8.t) : Tot bool =
  ((a +%^ b) |^ c) = d

// F*: (200+100) mod 256 = 44, 44 OR 0xF0 = 244
let lemma_add_or () : Lemma (add_or_compare 200uy 100uy 0xF0uy 244uy = true) = admit ()
```

**Generated C** (wrong):
```c
bool add_xor_compare(uint8_t a, uint8_t b, uint8_t c, uint8_t d) {
  return ((uint32_t)a + (uint32_t)b ^ (uint32_t)c) == d;
  // 300 ^ 0xFF = 0x12C ^ 0xFF = 0x1D3 = 467
  // 467 == 211 → false (WRONG, F* says true)
}

bool add_or_compare(uint8_t a, uint8_t b, uint8_t c, uint8_t d) {
  return ((uint32_t)a + (uint32_t)b | (uint32_t)c) == d;
  // 300 | 0xF0 = 0x12C | 0xF0 = 0x1FC = 508
  // 508 == 244 → false (WRONG, F* says true)
}
```

## Root Cause

[...]

Issue #692, opened 2026-03-23
[DeepTest] UInt16 has identical masking bugs as UInt8 (comparison, BNot, BShiftL)


elefthei:
## Bug Description

All UInt8 masking bugs (comparison after wrapping arithmetic, BNot, BShiftL) equally affect UInt16. The `mask` function at `lib/AstToCStar.ml:305-309` uses `0xFFFF` for UInt16, but masking is only applied before div/mod/shift — not before comparisons, and not after BNot/BShiftL.

## Reproduction

**F\* code** (verifies successfully):
```fstar
open FStar.UInt16

let add_then_compare (a b c: UInt16.t) : Tot bool =
  (a +%^ b) = c

let lemma_add_eq () : Lemma (add_then_compare 65535us 2us 1us = true) = ()

let sub_then_compare (a b c: UInt16.t) : Tot bool =
  (a -%^ b) = c

let lemma_sub_eq () : Lemma (sub_then_compare 0us 1us 65535us = true) = ()
```

**Generated C** (wrong):
```c
bool add_then_compare(uint16_t a, uint16_t b, uint16_t c) {
  return (uint32_t)a + (uint32_t)b == c;
  // 65535 + 2 = 65537, 65537 == 1 → false (WRONG)
}

bool sub_then_compare(uint16_t a, uint16_t b, uint16_t c) {
  return (uint32_t)a - (uint32_t)b == c;
  // 0 - 1 = 4294967295, 4294967295 == 65535 → false (WRONG)
}
```

## Affected Operations
- `65535us +%^ 2us = 1us` → F\*: `true`, C: `false`
- `0us -%^ 1us = 65535us` → F\*: `true`, C: `false`
- `256us *%^ 300us = 11264us` → F\*: `true`, C: `false`

## Root Cause

Same code paths as UInt8 bugs. `mk_arith` and `is_integer_arith` treat UInt8 and UInt16 identically — both are upcast to UInt32, and masking gaps affect both.

## Related Issues

Same root cause as the UInt8 comparison, BNot, and BShiftL bugs. A single fix in `mk_arith` addressing the masking strategy would resolve all of these for both widths.

Issue #691, opened 2026-03-23
[DeepTest] UInt8 left shift (BShiftL) result not masked — exceeds 8-bit width


elefthei:
## Bug Description

`BShiftL` (left shift) is in the "no mask" category of `mk_arith` (`lib/AstToCStar.ml:339`). When a UInt8 value is shifted left, the result can exceed 8 bits (e.g., `0x80 << 1 = 256`), but it is never masked back to 8 bits.

In F\*, `UInt8.shift_left` is defined as `(v * pow2 s) mod 256`, so bits shifted past the 8th position are lost. The C code retains them.

## Reproduction

**F\* code** (verifies successfully):
```fstar
open FStar.UInt8

let shift_then_compare (a: UInt8.t) (s: FStar.UInt32.t{FStar.UInt32.v s < 8}) (c: UInt8.t) : Tot bool =
  (a <<^ s) = c

// F*: 0x80 <<^ 1 = (128 * 2) mod 256 = 0
let lemma_shl_overflow () : Lemma (shift_then_compare 0x80uy 1ul 0uy = true) = ()
```

**Generated C** (wrong):
```c
bool shift_then_compare(uint8_t a, uint32_t s, uint8_t c) {
  return (uint32_t)a << s == c;
  // (uint32_t)0x80 << 1 = 256, 256 == 0 → false (WRONG)
}
```

## Affected Operations
- `0x80uy <<^ 1ul = 0uy` → F\*: `true`, C: `false`
- `(0x80uy <<^ 1ul) +%^ 5uy = 5uy` → F\*: `true`, C: `false`
- Chained: `(a <<^ s) +%^ b` retains extra bits in subsequent operations

## Root Cause

`BShiftL` at `lib/AstToCStar.ml:339` is in the "no mask" branch alongside Add/Sub/Mult:
```ocaml
| Add | AddW | Sub | SubW | Mult | MultW | BOr | BAnd | BXor | BNot | BShiftL ->
    CStar.Call (Op op, [ e1; e2 ])
```

## Suggested Fix

Add masking after `BShiftL` for UInt8/UInt16, similar to how `BShiftR` masks its left operand (line 344).

PR #703, opened 2026-03-28
Fix UInt8/UInt16 comparison after wrapping arithmetic (#694)

### LINKED ISSUE #689: [DeepTest] UInt8/UInt16 wrapping arithmetic + comparison produces wrong C code (missing mask)
## Bug Description

Karamel's `mk_arith` function (`lib/AstToCStar.ml:332-360`) upcasts UInt8/UInt16 operands to UInt32 for arithmetic. It correctly masks (`& 0xFF`) before div/mod/shift operations, but **comparison operators (`==`, `<`, `>`, `<=`, `>=`, `!=`) are not in `is_integer_arith`** (line 311-317), so no masking occurs before comparisons.

This causes verified F* code to produce **wrong boolean results** in extracted C.

## Reproduction

**F\* code** (verifies successfully):
```fstar
open FStar.UInt8

let sub_then_compare_eq (a b c: UInt8.t) : Tot bool =
  (a -%^ b) = c

let lemma_sub_eq () : Lemma (sub_then_compare_eq 0uy 1uy 255uy = true) = ()
```

**Generated C** (wrong):
```c
bool sub_then_compare_eq(uint8_t a, uint8_t b, uint8_t c) {
  return (uint32_t)a - (uint32_t)b == c;
  // When a=0, b=1: 4294967295 == 255 → false (WRONG, F* says true)
}
```

**Correct C** should be:
```c
return (uint8_t)((uint32_t)a - (uint32_t)b) == c;
```

## Affected Operations
All comparison operators on UInt8/UInt16 after wrapping arithmetic:
- `(0uy -%^ 1uy) = 255uy` → F\*: `true`, C: `false`
- `(255uy +%^ 2uy) = 1uy` → F\*: `true`, C: `false`
- `(200uy *%^ 200uy) = 64uy` → F\*: `true`, C: `false`
- `(200uy +%^ 100uy) <^ 50uy` → F\*: `true`, C: `false`

## Root Cause

`is_integer_arith` at line 311-317 of `lib/AstToCStar.ml` does not include `Eq`, `Neq`, `Lt`, `Lte`, `Gt`, `Gte`. Comparisons fall through to generic `EApp` (line 412) which does not call `mk_arith`, so the uint32 intermediate is compared directly without masking.

## Workaround

Bind the arithmetic result to a variable before comparing:
```fstar
let ok (a b c: UInt8.t) : bool =
  let result = a -%^ b in   // C assigns to uint8_t, truncating
  result = c                 // comparison on truncated value works
```

## Suggested Fix

Insert a narrowing cast/mask when a widened UInt8/UInt16 arithmetic expression appears as an operand of a comparison operator.

### LINKED ISSUE #694: [DeepTest] UInt8/UInt16 comparison bug propagates into if-conditions — wrong branch taken
## Bug Description

The UInt8/UInt16 comparison masking bug propagates into if-conditions, causing verified F\* code to **take the wrong branch** at runtime. This is the most dangerous manifestation of the missing-mask bug because it affects control flow, not just boolean return values.

If-conditions go through `mk_expr` (line 609 of `lib/AstToCStar.ml`), not `mk_arith`, so the comparison operands are not masked.

## Reproduction

**F\* code** (verifies successfully):
```fstar
open FStar.UInt8

let if_after_add (a b c: UInt8.t) : Tot FStar.UInt32.t =
  if (a +%^ b) = c then 1ul else 0ul

// F*: (255 +%^ 2) = 1, 1 = 1 → then-branch → returns 1
let lemma_if_add () : Lemma (if_after_add 255uy 2uy 1uy = 1ul) = ()

let if_after_sub (a b c: UInt8.t) : Tot FStar.UInt32.t =
  if (a -%^ b) = c then 42ul else 0ul

// F*: (0 -%^ 1) = 255, 255 = 255 → then-branch → returns 42
let lemma_if_sub () : Lemma (if_after_sub 0uy 1uy 255uy = 42ul) = ()
```

**Generated C** (wrong branch taken):
```c
uint32_t if_after_add(uint8_t a, uint8_t b, uint8_t c) {
  if ((uint32_t)a + (uint32_t)b == c)  // 257 == 1 → false!
    return 1U;
  else
    return 0U;  // WRONG: takes else-branch, returns 0
}

uint32_t if_after_sub(uint8_t a, uint8_t b, uint8_t c) {
  if ((uint32_t)a - (uint32_t)b == c)  // 4294967295 == 255 → false!
    return 42U;
  else
    return 0U;  // WRONG: takes else-branch, returns 0
}
```

## Impact

| F\* expression | F\* branch | C expression | C branch | C result |
|----------------|-----------|-------------|---------|----------|
| `if (255+%^2)=1 then 1 else 0` | then → 1 | `if (257==1)` | else → 0 | **WRONG** |
| `if (0-%^1)=255 then 42 else 0` | then → 42 | `if (4294967295==255)` | else → 0 | **WRONG** |
| `if (200+%^100)<50 then 99 else 0` | then → 99 | `if (300<50)` | else → 0 | **WRONG** |

## Security Impact

Wrong branch selection in extracted C code could cause:
- Incorrect MAC verification in cryptographic protocols
- Wrong key schedule computations
- Bypassed bounds checks on byte-level data
- Wrong state machine transitions

## Root Cause

If-condition expressions at `lib/AstToCStar.ml:609` use `mk_expr` directly, which does not apply `mk_arith` masking. The condition inherits the comparison bug from the generic `EApp` path (line 412).

## Related Issues

This is a consequence of the same root cause as the UInt8/UInt16 comparison masking bug — `is_integer_arith` does not include comparison operators. Fixing that issue would also fix this one.

elefthei:
Comparisons (Eq, Neq, Lt, Lte, Gt, Gte) on UInt8/UInt16 after wrapping arithmetic (add_mod, sub_mod, mul_mod, shift_left, xor, or, etc.) produced wrong C code. The uint32 intermediate from mk_arith was compared directly without masking back to the original width.

For example, (255uy +%^ 1uy) = 0uy generated:
  (uint32_t)255 + (uint32_t)1 == 0  =>  256 == 0  =>  false (WRONG)

Fix: add comparison operators to is_integer_arith so they enter mk_arith, and mask both operands before comparing (same strategy as Div/Mod).

Now generates:
  ((uint32_t)255 + (uint32_t)1 & 0xFFU) == (uint32_t)0  =>  0 == 0  =>  true

Fixes #694. Also fixes #689, #691, #692, #693 (same root cause).

### DIFF
--- a/lib/AstToCStar.ml
+++ b/lib/AstToCStar.ml
@@ -311,7 +311,8 @@ and mask w e =
 and is_integer_arith op w =
   w <> K.Bool && not (Constant.is_float w) && K.is_unsigned w && match op with
   | K.Add | AddW | Sub | SubW | Div | DivW | Mult | MultW | Mod
-  | BOr | BAnd | BXor | BShiftL | BShiftR | BNot ->
+  | BOr | BAnd | BXor | BShiftL | BShiftR | BNot
+  | Eq | Neq | Lt | Lte | Gt | Gte ->
       true
   | _ ->
       false
@@ -342,6 +343,8 @@ and mk_arith env e =
           CStar.Call (Op op, [ mask_if a1 w e1; mask_if a2 w e2 ])
       | BShiftR ->
           CStar.Call (Op op, [ mask_if a1 w e1; e2 ])
+      | Eq | Neq | Lt | Lte | Gt | Gte ->
+          CStar.Call (Op op, [ mask_if a1 w e1; mask_if a2 w e2 ])
       | _ ->
           assert false
       end, false, w1 || w2
--- a/test/SmallIntPromotion.fst
+++ b/test/SmallIntPromotion.fst
@@ -1,0 +1,27 @@
+/// Regression test for karamel issue #694.
+/// Tests that UInt8/UInt16 integer promotion correctly masks results
+/// before comparisons.
+module SmallIntPromotion
+
+open FStar.UInt8
+
+/// Comparison after wrapping arithmetic must mask to 8 bits.
+/// Use runtime parameters to prevent constant folding.
+let check_add_compare (a b c: UInt8.t) : bool = (a +%^ b) = c
+let check_sub_compare (a b c: UInt8.t) : bool = (a -%^ b) = c
+let check_add_lt (a b c: UInt8.t) : bool = (a +%^ b) <^ c
+let check_shl_compare (a: UInt8.t) (s: FStar.UInt32.t{FStar.UInt32.v s < 8}) (c: UInt8.t) : bool =
+  (a <<^ s) = c
+
+/// Comparison in if-condition must also mask.
+let check_if_add (a b c: UInt8.t) : FStar.Int32.t =
+  if (a +%^ b) = c then 0l else 1l
+
+let main () : FStar.Int32.t =
+  if check_add_compare 255uy 2uy 1uy
+  && check_sub_compare 0uy 1uy 255uy
+  && check_add_lt 200uy 100uy 50uy
+  && check_shl_compare 0x80uy 1ul 0uy
+  && check_if_add 255uy 2uy 1uy = 0l
+  then 0l
+  else 1l


elefthei:
@protz need to merge this first as #704 CI fails because of this bug.

PR #721, opened 2026-04-06
Add `-fhoist-locals`, `-finitialize-locals`

tahina-pro:
tl;dr: This PR adds two passes, along with corresponding tests:

* `-fhoist-locals`: hoist local variables to the beginning of functions
* `-finitialize-locals`: add initializers to uninitialized local variables

You can find an example use of `-fhoist-locals` in my EverParse fork: tahina-pro/quackyducky@61a825fa370f42f197fd1424395634e6a809b7e7

**Please do not rebase this branch** to a commit that is not on `everparse-fstar1`, which is the Karamel branch (an ancestor of `master`) that EverParse `master` is currently using.

# `-fhoist-locals`

This option hoists local variables to the beginning of functions:
* Declarations already present at the beginning of functions are unchanged. Elsewhere:
* Declarations without initializers are hoisted to the beginning of functions
* Declarations with initializers are split into two parts: a declaration without initializer hoisted at the beginning of functions, and initializers become assignments:
  * local arrays with non-constant size, or with per-element initializers make hoisting stop: Karamel emits a warning, by default as an error
  * local arrays with a single initializer (EBufCreate not Any) are split into a declaration without initializer and a EBufFill operation

# `-finitialize-locals` 

This option adds initializers to uninitialized local variables.

It takes an argument:
* `no`: do not initialize uninitialized local variables. This is the default.
* `c23`: initialize uninitialized local variables with `= {}`. This is enough to initialize everything without descending into nested structs, etc.
* `c99`: deep-initialize structs, using a field designator (`.field =`) to initialize each struct field. This option descends into nested structs, and initializes each integer with 0 and each pointer with NULL.
* `c89`: same as `c99`, but without the field designator syntax

`c99` and `c89` initialize unions by initializing their first field, which is the only possible way to initialize a union in C89.

This option emits a warning, by default an error, if it encounters an unknown (abstract) type.

Other than for integer fields, this option *does not* rely on the old GCC `= {0}` behavior, which has been removed in recent (>= 14.1) versions of GCC.


tahina-pro:
Sorry for my late reply, I am just back from vacation.
This PR is AI co-authored (the commit message is tagged as such).
I have been testing this PR with EverParse and it seems to work well.

protz (changes requested):
Hi Tahina, thanks for the PR. These are high-level comments before I review this more in detail.

- Can you explain why you chose to duplicate the existing use analysis? There is already an infrastructure in place to perform use-analysis (lib/UseAnalysis.ml), and this already tries to handle the case of the if-then-else. There is also a MaybeUnused flag that already appears in places and emits the correct macro. It looks like you chose to circumvent all of this existing infrastructure and write fresh code that operates at the very end of the pipeline on a syntactic tree that is only intended for pretty-printing. **Request:** please minimize the issue and explain why the current code doesn't work for you. Furthermore, please fix the existing use-analysis.
- For the hoisting of locals, two remarks.
  1. Can we drop C89 compatibility mode? Frankly, C89 is nearly 30 years old and the C89 mode has at least one bad bug that I know of in krml. This mode is a liability and I would like to eliminate it. Can we ditch it instead of adding more support for it?
  2. You are duplicating (or perhaps, your LLM is) a lot of code in the existing hoisting code. Why not change the behavior of the existing hoisting phase with a flag that percolates the let-bindings further up? (in Simplify.ml I believe) **Request:** convince me that adding a whole lot of fresh code rather than work with the existing code makes sense, or tweak the existing hoist phase that is already written to operate the way you want.
- The `initialize_locals` phase should operate at the level of CStar, at the very least. Like I wrote in the other PR, C11 is only a syntactic tree and manipulating it is a nightmare because of the spec-and-decl convoluted representation of C (not to mention the types...). **Request:** please revise this.

Generally, I am under the impression that your PR is either working around existing code rather than fixing it, and sometimes operating at the wrong level of abstraction. Let's revise this to make sure we don't add 700 lines of fresh code unless there's a very good reason to do so.

Perhaps a good way to proceed would be to break this down into individual PRs. It would be easier to review.

Thanks.

protz:
Edit: there is already `-fc89-scope` and a corresponding implementation. Why did you choose to reimplement this logic from scratch?

PR #730, opened 2026-04-07
Fix an edge case in visibility analysis

protz:
A constant's body may contain references to other declarations (e.g. function calls). This is weird, but is technically possible via all frontends, and was recently found when adding ML-DSA to Eurydice's test suite.


protz:
The diff in `eurydice` is being refreshed here: https://github.com/AeneasVerif/eurydice/pull/394

PR #734, opened 2026-04-29
Makefile: allow skipping krmllib build when LOWSTAR=false

mtzguido:
When using newer F*'s without Low* support, the default makefile rule will fail since it tries to build krmllib. This patch allows to set LOWSTAR=false in the environment to avoid it, but still get a working local install (just without krmllib checked files).

(This patch is already in the fstar2 branch that the F* submodule tracks.)


protz (approved):

protz:
Makes sense, thanks

PR #735, opened 2026-04-30
Add constant folding for booleans

mtzguido:
We sometimes generate trivial boolean expressions after inlining. This patch makes them simplify. A test is added.

Example output before:

	void BoolFold_main(void)
	{
	  BoolFold_f(true && true);
	  BoolFold_f(true && !true);
	  BoolFold_f(true || true);
	  BoolFold_f(true || !true);
	  BoolFold_f(!true && true);
	  BoolFold_f(!true && !true);
	  BoolFold_f(!true || true);
	  BoolFold_f(!true || !true);
	}

After:

	void BoolFold_main(void)
	{
	  BoolFold_f(true);
	  BoolFold_f(false);
	  BoolFold_f(true);
	  BoolFold_f(true);
	  BoolFold_f(false);
	  BoolFold_f(false);
	  BoolFold_f(true);
	  BoolFold_f(false);
	}

This does not generate a diff in krmllib dist.

--

Hi Jonathan. At a high level I wonder if you agree that beefing up these simplification passes is worth it to generate higher-quality code, since it is of course more TCB. Happy to gate this via an option if you think that's better. See also upcoming PR on ternary operator.

### DIFF
--- a/lib/Simplify.ml
+++ b/lib/Simplify.ml
@@ -507,6 +507,35 @@ let constant_fold = object (self)
         | _ -> EApp (e, [ e1; e2 ])
     )
 
+    | EOp (K.And, _), [e1; e2] -> (
+        let e1 = self#visit_expr env e1 in
+        let e2 = self#visit_expr env e2 in
+        match e1.node, e2.node with
+        | EBool false, _ -> EBool false
+        | _, EBool false -> EBool false
+        | EBool true, e2 -> e2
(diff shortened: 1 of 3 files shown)


protz (approved):
I am happy with this. We know that a lot of code fed into krml is meta-programmed, and as such, working a little bit harder is worth the improvement in code quality. Thank you!

PR #737, opened 2026-04-30
-fternary: Support generating C's ternary operator (?:)

mtzguido:
This automatically translates if-then-else to the ternary operator when
the arguments are pure expressions. `ETernary` is added as an AST node,
and a pass over the AST tries to turn some `EIfThenElse` (as long as the
arguments are pure and without intermediate statements) into `ETernary`,
which get printed as C's `?:`. This is implemented here and not in, say,
AstToCStar, since by that point the `EIfThenElse` nodes have been
flattened into blocks and assignments to intermediate variables, so the
structure is lost.

Requiring the arguments to be pure is just a heuristic and not need. The
requirement for no intermediate definitions is meant to prevent trying
to use a ternary op for something like `f () && (let x = g() in h x)`,
since the RHS must be a block.

When generating Rust or CFlat, `ETernary` nodes are turned back into
`EIfThenElse`, reversing the heuristic detection.

The flag defaults to false, but making it default to true would pass the
test suite (modulo some output changing). Below is an excerpt of the
diff that we would generate if this flag became the default. Some of it
is quite desirable, particularly as the ternary ops fully disappear and
we just end up simplifying blocks away!

AI disclaimer: This was mostly hand written, just with some help from
Claude to figure out what to do about #ifdef detection and having it add
a few tests.

This is based on the constant folding of bools since that becomes more
important after this patch. (Though not strictly necessary.)

```diff
diff --git a/krmllib/dist/generic/FStar_Attributes.c b/krmllib/dist/generic/FStar_Attributes.c
index https://github.com/FStarLang/karamel/commit/2a2d2603a0193b9bfee920d4c2024111b5664566..678323ce 100644
--- a/krmllib/dist/generic/FStar_Attributes.c
+++ b/krmllib/dist/generic/FStar_Attributes.c
@@ -11,14 +11,7 @@ FStar_Attributes_uu___is_PpxDerivingShow(
   FStar_Attributes___internal_ocaml_attributes projectee
 )
 {
-  if (projectee.tag == FStar_Attributes_PpxDerivingShow)
-  {
-    return true;
-  }
-  else
-  {
-    return false;
-  }
+  return projectee.tag == FStar_Attributes_PpxDerivingShow;
 }

 bool
diff --git a/krmllib/dist/generic/FStar_Order.c b/krmllib/dist/generic/FStar_Order.c
index https://github.com/FStarLang/karamel/commit/04862dffce69ab63914f19a194489a7e4fbb4751..2eeab533 100644
--- a/krmllib/dist/generic/FStar_Order.c
+++ b/krmllib/dist/generic/FStar_Order.c
@@ -113,13 +113,9 @@ FStar_Order_order FStar_Order_order_from_int(krml_checked_int_t i)
   {
     return FStar_Order_Lt;
   }
-  else if (i == (krml_checked_int_t)0)
-  {
-    return FStar_Order_Eq;
-  }
   else
   {
-    return FStar_Order_Gt;
+    return i == (krml_checked_int_t)0 ? FStar_Order_Eq : FStar_Order_Gt;
   }
 }

diff --git a/krmllib/dist/generic/FStar_UInt128_Verified.h b/krmllib/dist/generic/FStar_UInt128_Verified.h
index https://github.com/FStarLang/karamel/commit/7aa2a6d83f9c4b922f65a310c59c60d8147644f7..1f3e2694 100644
--- a/krmllib/dist/generic/FStar_UInt128_Verified.h
+++ b/krmllib/dist/generic/FStar_UInt128_Verified.h
@@ -159,14 +159,9 @@ FStar_UInt128_shift_left_large(FStar_UInt128_uint128 a, uint32_t s)
 static inline FStar_UInt128_uint128
 FStar_UInt128_shift_left(FStar_UInt128_uint128 a, uint32_t s)
 {
-  if (s < FStar_UInt128_u32_64)
-  {
-    return FStar_UInt128_shift_left_small(a, s);
-  }
-  else
-  {
-    return FStar_UInt128_shift_left_large(a, s);
-  }
+  return
+    s < FStar_UInt128_u32_64 ? FStar_UInt128_shift_left_small(a, s) : FStar_UInt128_shift_left_large(a,
+      s);
 }

 static inline uint64_t FStar_UInt128_add_u64_shift_right(uint64_t hi, uint64_t lo, uint32_t s)
diff --git a/krmllib/dist/generic/WasmSupport.c b/krmllib/dist/generic/WasmSupport.c
index https://github.com/FStarLang/karamel/commit/989ee03c84013c7d7536f34db84402e88a3ee783..e8a37158 100644
--- a/krmllib/dist/generic/WasmSupport.c
+++ b/krmllib/dist/generic/WasmSupport.c
@@ -8,14 +8,7 @@

 uint32_t WasmSupport_align_64(uint32_t x)
 {
-  if (!((x & 0x07U) == 0U))
-  {
-    return (x & 4294967288U) + 0x08U;
-  }
-  else
-  {
-    return x;
-  }
+  return !((x & 0x07U) == 0U) ? (x & 4294967288U) + 0x08U : x;
 }

 void WasmSupport_check_buffer_size(uint32_t s)
diff --git a/test/pulse/Break.expected.c b/test/pulse/Break.expected.c
index https://github.com/FStarLang/karamel/commit/979c520eb8baca83aadb9cdb1af0c81dc158a7e5..f456a21c 100644
--- a/test/pulse/Break.expected.c
+++ b/test/pulse/Break.expected.c
@@ -7,21 +7,8 @@ void Break_simple_break(void)
 {
   bool k = true;
   bool _break = false;
-  bool cond;
-  if (_break)
-    cond = false;
-  else
-    cond = k;
-  while (cond)
-  {
+  while (!_break && k)
     _break = true;
-    bool ite;
-    if (_break)
-      ite = false;
-    else
-      ite = k;
-    cond = ite;
-  }
 }

 void Break_break_continue_and_return(uint8_t which)
diff --git a/test/pulse/Example_Hashtable.expected.c b/test/pulse/Example_Hashtable.expected.c
index https://github.com/FStarLang/karamel/commit/353fbbf99e157f3c5dcdaac3861f713e975335df..8b2f92d4 100644
--- a/test/pulse/Example_Hashtable.expected.c
+++ b/test/pulse/Example_Hashtable.expected.c
@@ -248,12 +248,7 @@ insert__size_t_Example_Hashtable_data(
   size_t off = (size_t)0U;
   size_t idx = (size_t)0U;
   bool _break = false;
-  bool cond;
-  if (_break)
-    cond = false;
-  else
-    cond = true;
-  while (cond)
+  while (!_break)
   {
     size_t voff = off;
     size_t sum = cidx + voff;
@@ -334,12 +329,6 @@ insert__size_t_Example_Hashtable_data(
         "unreachable (pattern matches are exhaustive in F*)");
       KRML_HOST_EXIT(255U);
     }
-    bool ite;
-    if (_break)
-      ite = false;
-    else
-      ite = true;
-    cond = ite;
   }
   size_t __anf0 = idx;
   write_ref__Pulse_Lib_HashTable_Spec_cell_size_t_Example_Hashtable_data(&contents,
diff --git a/test/pulse/Null.expected.c b/test/pulse/Null.expected.c
index https://github.com/FStarLang/karamel/commit/3244f62de58ac482496a8be3b15ce6f0fc94aa1a..d3711b80 100644
--- a/test/pulse/Null.expected.c
+++ b/test/pulse/Null.expected.c
@@ -17,9 +17,6 @@ krml_checked_int_t *Null_foo(void)

 krml_checked_int_t Null_test(krml_checked_int_t *x1)
 {
-  if (x1 == NULL)
-    return (krml_checked_int_t)0;
-  else
-    return *x1;
+  return x1 == NULL ? (krml_checked_int_t)0 : *x1;
 }
```

### DIFF
--- a/lib/PrintAst.ml
+++ b/lib/PrintAst.ml
@@ -369,6 +369,12 @@ and print_expr env { node; typ; meta } =
       string "NULL" ^^ langle ^^ print_typ typ ^^ rangle
   | EContinue ->
       string "continue"
+  | ETernary (c, t, e) ->
+      parens_with_nesting (
+        print_expr env c ^/^
+          string "?" ^/^ print_expr env t ^/^
+          string ":" ^/^ print_expr env e
+      )
 
 and print_poly_comp = function
   | PEq -> equals
--- a/lib/PrintC.ml
+++ b/lib/PrintC.ml
@@ -344,6 +344,16 @@ and p_expr' curr = function
       p_stmts stmts
   | CxxInitializerList init ->
       p_init init
+  | Ternary (c, t, e) ->
+      let mine = 13 in
+      let c = p_expr' 12 c in
+      (* "The expression in the middle of the conditional operator (between ?
+      and :) is parsed as if parenthesized: its precedence relative to ?: is
+      ignored." Hence the 15 (max) here. *)
+      let t = p_expr' 15 t in
+      let e = p_expr' 13 e in
+      paren_if curr mine @@
+      group c ^^ space ^^ align (qmark ^^ space ^^ group t ^/^ colon ^^ space ^^ group e)
 
 (* statement-level comment *)
 and p_comment s =
--- a/lib/Simplify.ml
+++ b/lib/Simplify.ml
@@ -526,7 +526,7 @@ let constant_fold = object (self)
         | _, EBool true -> EBool true
         | EBool false, e2 -> e2
         | e1, EBool false -> e1
-        | _ -> EApp (e, [e1;e2 ])
+        | _ -> EApp (e, [e1; e2])
     )
 
     | EOp (K.Not, _), [e1] -> (
@@ -538,6 +538,44 @@ let constant_fold = object (self)
 
     | _ ->
         EApp (self#visit_expr env e, List.map (self#visit_expr env) es)
+
+  method! visit_ETernary env c t e =
+    let c = self#visit_expr env c in
+    let t = self#visit_expr env t in
+    let e = self#visit_expr env e in
+    assert (c.typ = TBool);
+    (* Some of the rules below are gated with type information, to make sure we
+    don't end up rewriting something like `p ? true : 42` into `p || 42`, which
+    would booleanize the 42 into 1. Karamel's typechecker should prevent this
+    from being the case (the types of the branches should match and be either
+    TBool or TInt), but better safe than sorry. *)
+    match c.node, t.node, e.node with
+    | EBool true, _, _ ->
+      (* true ? t : e  ~>  t *)
+      t.node
+    | EBool false, _, _ ->
+      (* false ? t : e  ~>  e *)
+      e.node
+    | _, EBool false, EBool true ->
+      (* c ? false : true  ~>  !c *)
+      (mk_not c).node
+    | _, EBool true, EBool false ->
+      (* c ? true : false  ~>  c. Note, we know that c is TBool. *)
+      c.node
+    | _, EBool true, _ when e.typ = TBool ->
+      (* c ? true : e  ~>  c || e *)
+      (mk_or c e).node
+    | _, EBool false, _ when e.typ = TBool ->
+      (* c ? false : e  ~>  !c && e *)
+      (mk_and (mk_not c) e).node
+    | _, _, EBool true when t.typ = TBool ->
+      (* c ? t : true  ~>  !c || t *)
+      (mk_or (mk_not c) t).node
+    | _, _, EBool false when t.typ = TBool ->
+      (* c ? t : false  ~>  c && t *)
+      (mk_and c t).node
+    | _ ->
+      ETernary (c, t, e)
 end
 
 let remove_ignore_unit = object
@@ -672,6 +710,40 @@ let let_if_to_assign = object (self)
 
     | _ ->
         EAssign (e1, e2)
+end
+
+(* Check if an expression references any ifdef variable. If so, we must keep it
+   as EIfThenElse so that AstToCStar can compile it to #ifdef. *)
+let references_ifdef ifdefs =
+  (object
+    inherit [_] reduce
+    method private zero = false
+    method private plus = (||)
+    method! visit_EQualified _ lid = Idents.LidSet.mem lid ifdefs
+  end)#visit_expr_w ()
+
+(* Eligible for use inside a C ternary: readonly (no side effects, no function
+   calls) AND no statement constructs (ELet, EIfThenElse) anywhere in the
+   subtree. *)
+let is_ternary_eligible = (object
+  inherit [_] readonly_visitor
+  method! visit_ELet _ _ _ _ = false
+  method! visit_EIfThenElse _ _ _ _ = false
+end)#visit_expr_w ()
+
+let use_ternary_op ifdefs = object (self)
+  inherit [_] map
+
+  method! visit_EIfThenElse env c t e =
+    let c = self#visit_expr env c in
+    let t = self#visit_expr env t in
+    let e = self#visit_expr env e in
+    if is_ternary_eligible c &&
+       is_ternary_eligible t &&
+       is_ternary_eligible e &&
+       not (references_ifdef ifdefs c)
+    then ETernary (c, t, e)
+    else EIfThenElse (c, t, e)
 end
 
 (* Transform functional updates of the form x.(i) <- { x.(i) with f = e } into
@@ -1565,6 +1637,14 @@ and hoist_expr tbl loc pos e =
       else
         Warn.fatal_error "%a: %s" Loc.ploc loc "[continue] expressions should only appear in statement position"
 
+  | ETernary (e1, e2, e3) ->
+      let lhs1, e1 = hoist_expr loc Unspecified e1 in
+      let lhs2, e2 = hoist_expr loc UnderConditional e2 in
+      let e2 = nest lhs2 e2.typ e2 in
+      let lhs3, e3 = hoist_expr loc UnderConditional e3 in
+      let e3 = nest lhs3 e3.typ e3 in
+      lhs1, mk (ETernary (e1, e2, e3))
+
 (* TODO: figure out if we want to ignore the other cases for performance
  * reasons. *)
 class hoist = object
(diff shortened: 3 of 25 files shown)


lib/PrintAst.ml
```diff
@@ -369,6 +369,10 @@ and print_expr env { node; typ; meta } =
       string "NULL" ^^ langle ^^ print_typ typ ^^ rangle
   | EContinue ->
       string "continue"
+  | ETernary (c, t, e) ->
+      print_expr env c ^^ space ^^ string "?" ^^ space ^^
```
protz: almost certainly need a group here

mtzguido:
Some comments:
- It's probably worth a skeptic review
- The rewrite rules to constant fold ternary operators all preserve the semantics even if c,t,e may be impure. But I am relying on some typing to rewrite things like `c ? true : false` into `c`, instead of `!!c`. Also rewriting `c ? true : e` to `c || e`, relying on the fact that `e` must be of type Bool. Confirmation that this makes sense would be good.
- I also wrote those rules using `is_true` and `is_false` since I though I had to detect `EBool true` and also `EConstant (Bool, _some_nonzero)`, but AFAICT the latter don't seem to ever occur. I would probably rewrite this to use a pattern match if my understanding is true.

lib/PrintC.ml
```diff
@@ -344,6 +344,12 @@ and p_expr' curr = function
       p_stmts stmts
   | CxxInitializerList init ->
       p_init init
+  | Ternary (c, t, e) ->
+      let mine = 13 in
+      let c = p_expr' 12 c in
+      let t = p_expr' 15 t in
```
protz, line 353: I assume you are relying on this note:

> The expression in the middle of the conditional operator (between ? and :) is parsed as if parenthesized: its precedence relative to ?: is ignored. 

here https://en.cppreference.com/c/language/operator_precedence ?

if so, put a comment, because it's a weird case that doesn't follow the pattern of the other cases

lib/Simplify.ml
```diff
@@ -415,6 +415,18 @@ let wrapping_arithmetic = object (self)
         EApp (self#visit_expr env e, List.map (self#visit_expr env) es)
 end
 
+(* Why is there EBool but also EConstant with Bool width?
+   Should we also detect EConstants? *)
+let is_true (t : expr) : bool =
```
protz: this is a historical mistake that I would be more than happy to see fixed

I think the invariant is that there are no EConstants at bool type and they are EBool instead, but maybe adding  pass that enforces this in debug mode would be a good start

lib/Simplify.ml
```diff
@@ -507,8 +519,75 @@ let constant_fold = object (self)
...
+      (* true ? t : e  ~>  t *)
+      t.node
+    else if is_false c then
+      (* false ? t : e  ~>  e *)
+      e.node
+    else if is_false t && is_true e then (
+      (* c ? false : true  ~>  !c *)
+      (mk_not c).node
+    ) else if is_true t && is_false e then (
+      (* NOTE: this rewriting and the ones below rely on well-typing of these
```
protz: ocnfirm that this is enforced in Checker?

protz (changes requested):
This is great and something I've wanted for a long while. I would suggest we make this the default since it generates obviously better code. I would  even be in favor of removing the option.

I have a few minor comments but apart from that it looks great. Thanks!

protz:
I answered both of your questions in inline comments I think, but just to be exhaustive:
- re ternary-as-bool: let's enforce this in Checker. It's ok to have this restriction as long as krml enforces it via the Checker
- re EConstant (Bool, _) -> historical mistake, happy to see it fixed (e.g. by a llm-driven rewrite) to clean up that unfortunate mistake

we have good code coverage now so we should be able to detect changes to the generated code

my only concern is HACL* -- how do we check the impact of such changes on HACL*? is HACL* stuck on an old krml version or is still picking up new krml versions? ping @R1kM

mtzguido:
Perfect, thanks for reviewing! I'll submit an update probably today.

lib/Simplify.ml
mtzguido: The checker does enforce that the condition is a Bool, and that the two branches match in type. I guess I'm relying on the fact that if `t = EBool _`, then the type of `t` (and therefore the whole expression) is `TBool`. That's probably true, but maybe we can gate the rewrite on whether the type is `TBool`, which should always hold but give some stopgap in case things go wrong?

mtzguido:
OK I pushed some updates:
- Made it the default (and updated dist + expected.c files)
- Gated the simplifications with typing info. They all kick in anyway (which is a good sign) and probably safer.
- Added a note about the precedence
- Tweaked printing, things look nicer.

I think you can skim through the new commits in order to review. I'll squash everything before merging.

Also curious to see the diff in eurydice when CI runs.

mtzguido:
> * re EConstant (Bool, _) -> historical mistake, happy to see it fixed (e.g. by a llm-driven rewrite) to clean up that unfortunate mistake

This is a bit of a mess since the right thing (IMO) is removing `Bool` from `Constant.width`, but this affects the ABI (it's used by `InputAst`). We could reject them during the InputAst->Ast conversion, but we'd still have to match them everywhere. Removing it implies patches to both "fstar1" and "fstar2".

An alternative is duplicating `Constant` into `InputConstant` and having Bool only in the latter...

lib/Simplify.ml
protz: it would be pretty bad if the type of the outer expression was TBool but the type of the then-branch or the else-branch was not TBool -- I think you can safely assume this is the case

protz:
> An alternative is duplicating `Constant` into `InputConstant` and having Bool only in the latter...

This seems reasonable.

This PR looks good to me -- let's run CI and assess the diff on Eurydice?

mtzguido:
> This seems reasonable.

Ack, I'll send a separate PR. I have the removal implemented already, just need the Input version.

> This PR looks good to me -- let's run CI and assess the diff on Eurydice?

I think Github is being stupid. I did see a green and no diff on Eurydice. I'll clean up the history and update the PR, maybe that makes it kick in.

Should I check HACL* locally? If there's a workflow to do it that's much easier, but I can also do it.

protz (approved):

PR #738, opened 2026-04-30
Remove Bool width

mtzguido:
Prompted from discussion [here](https://github.com/FStarLang/karamel/pull/737#issuecomment-4355516642).

Question: The op's from Assign onwards are not generated by F* (they are not even in its op type). Are they generated from elsewhere or just internal? Idem for Float types, though these will be added to F*.

### DIFF
--- a/CI.md
+++ b/CI.md
@@ -21,7 +21,7 @@ $ cd eurydice
 $ (cd karamel && git fetch && git checkout my-pr-branch && git merge)
 $ (cd out/ && git clean -fdx)
 $ make test -j
-$ nix flake update karamel --override-input karamel "github:fstarlang/karamel/$(cd karamel && git rev-parse head)"
+$ nix flake update karamel --override-input karamel "github:fstarlang/karamel/$(cd karamel && git rev-parse HEAD)"
 $ git commit -am "Update krml to latest revision on branch my-pr-branch"
 ```
 
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -214,7 +214,7 @@ type expr' =
   | EOpen of ident * atom_t
     (** [ident] for debugging purposes only *)
 
-  | EOp of op * width
+  | EOp of op * typ_wo
   | EQualified of lident
   | EConstant of constant
   | EUnit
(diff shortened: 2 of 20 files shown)


mtzguido:
```
       > File "lib/Builtin.ml", line 548, characters 49-54:
       > 548 |     let meta = mk_sizeT (EApp (Helpers.mk_op Sub SizeT, [ r_end; r_start ])) in
       >                                                        ^^^^^
       > Error: This variant expression is expected to have type Krml.Ast.typ_wo
       >        There is no constructor SizeT within type Krml.Ast.typ
       >
```
Ah right, Eurydice uses karamel as a library, it will need a patch.

protz:
> The op's from Assign onwards are not generated by F*

Generated from Scylla I think. I also thought that PreIncr and PostIncr were, at some point, synthesized via peephole optimizations, but that might be long-gone.

Nit: your rewrite uses `EBool` for a couple ASTs that rely on the type-disambiguation feature of OCaml and do not have an `E` prefix

The way to patch Eurydice is to submit a PR there that relies on this particular commit, refresh the diff and the krml dependency (`make nix-update-karamel`), then merge it so that this PR picks up Eurydice main with the diff already integrated.

mtzguido:
Eurydice PR: https://github.com/AeneasVerif/eurydice/pull/400



> Nit: your rewrite uses `EBool` for a couple ASTs that rely on the type-disambiguation feature of OCaml and do not have an `E` prefix

Not sure I get this, do you mean that they are lacking a module prefix? As in
```ocaml
  | EBool b ->
      locals, EBool b
```
in AstToCFlat? I do find that looks funny, I can qualify the names if you prefer.

mtzguido:
Ah sorry, I see what you meant. I renamed EBool to Bool in CFlat. In MiniRust, doing so gives me an error about `visit_Bool` that I don't quite understand, probably a clash between the Bool type and Bool expression?
```
File "lib/MiniRust.ml", lines 69-143, characters 0-39:
 69 | type array_expr =
 70 |   | List of expr list
 71 |   | Repeat of expr * expr (* [ e; n ] *)
 72 | [@@deriving show,
 73 |   visitors { variety = "map"; name = "map_expr" ;
...
140 |   atom: atom_t
141 | }
142 |
143 | and atom_t = Atom.t [@ visitors.opaque]
Error: The method visit_Bool has type 'a -> bool -> expr
       but is expected to have type 'env. 'env -> typ
       The universal variable 'env would escape its scope
```

protz:
yes there's an attribute somewhere to change the name of the generated visitor method -- there are other cases where this happens

mtzguido:
I see thanks! Updated.

PR #740, opened 2026-05-03
Update dist + add a daily workflow for it

mtzguido:
This will open a PR whenever there is a diff. CI should run on the PR (github disables that by default, but using DZOMO_GITHUB_TOKEN should make it kick in). The existing PR is updated if it is unmerged and the dist changed again.

Knowing actions, it's possible this fails in some weird way. I'll fix if so. It's hard to tell before deploying.


protz (approved):

protz:
Thanks!

PR #741, opened 2026-05-03
Fix UInt8/UInt16 masking: switch scrutinee, ETernary, ECast

### LINKED ISSUE #707: [DeepTest] UInt8/UInt16 switch scrutinee not masked after wrapping arithmetic
## Bug Description

**Category**: Wrong control flow (wrong branch taken)

When F* `match` on UInt8/UInt16 is compiled as a C `switch` statement, the scrutinee expression retains its widened uint32 value. The switch compares the uint32 value against uint8 case labels, taking the wrong branch when wrapping arithmetic overflows 8 bits.

## Expected vs Actual

```fstar
let via_switch (a b: UInt8.t) : UInt32.t =
  match (a +%^ b) with
  | 0uy -> 10ul
  | 1uy -> 20ul
  | _   -> 30ul
```

With `a=200, b=56`: F* computes `(200+56) mod 256 = 0`, matches `0uy`, returns `10`.

Generated C:
```c
switch ((uint32_t)a + (uint32_t)b) {  // 256, not 0!
  case 0U: return 10U;
  case 1U: return 20U;
  default: return 30U;              // takes this branch
}
```

C returns `30` (WRONG, F* says `10`).

## Why this is distinct from #694

Bug #694 fixed comparisons (`==`, `<`, etc.) by adding them to `mk_arith`. But the `switch` scrutinee path is different: `try_mk_switch` in `DataTypes.ml` puts the raw expression into `ESwitch`, bypassing `mk_arith` entirely. The if-else compilation path (`compile_match`) is correct because it binds the scrutinee to a `uint8_t` variable first.

## Root Cause

`DataTypes.ml` `try_mk_switch` extracts the scrutinee expression without masking. `AstToCStar.ml` defers masking for `Add`/`Sub`/`Mul` (relying on the consumer to truncate), but `switch` is not a truncating consumer.

## Impact

Any F* `match` on UInt8/UInt16 that involves wrapping arithmetic in the scrutinee may take the wrong branch in extracted C code.

mtzguido:
Three issues:
1- The ESwitch case passed the scrutinee through mk_expr, bypassing
   mk_arith masking.
2- mk_arith's catch-all marked ETernary as atomic, hiding non-atomic
   arithmetic branches from consumers (switch, comparisons).
3- ECast applied the cast to the widened uint32 value without masking,
   leaking overflow bits on widening casts.

Fix: for UInt8/UInt16 switch scrutinees, route through mk_arith and mask non-atomic results. Add an ETernary case to mk_arith that recurses into both branches, returning is_atomic only when both are atomic. For ECast, route UInt8/UInt16 sources through mk_arith and mask before casting.

Add SmallIntSwitch.fst, SmallIntTernary.fst, SmallIntCast.fst regression tests.

Fixes #707

--

Hi Jonathan, this is coauthored with Claude. It fixes #707 (which is legit) and a few other variants. The three tests fail before this patch. I'm not sure the style of the patch is that pleasant, since it seems easy to forget about this mk_arith dance and reintroduce variants of this bug later.

### DIFF
--- a/lib/AstToCStar.ml
+++ b/lib/AstToCStar.ml
@@ -363,6 +363,13 @@ and mk_arith env e =
          that every subexpression operates over unsigned int until the final
          cast, or until a mask is needed to preserve semantics. *)
       mk_expr env false false e, true, false (* C++: a constant that is wider than the intended type, but in an initializer list, is ok *)
+  | ETernary (c, t, f) ->
+      (* Recurse into branches so that non-atomic arithmetic is not hidden
+         behind a ternary marked as "atomic". *)
+      let c = mk_expr env false false c in
+      let t, a1, w1 = mk_arith env t in
+      let f, a2, w2 = mk_arith env f in
+      CStar.Ternary (c, t, f), a1 && a2, w1 || w2
   | _ ->
       (* Something else. Who knows? Maybe a function call, a field reference, a
          variable, a global... which will be upcast into `int`, which is *not*
@@ -454,7 +461,17 @@ and mk_expr env in_stmt under_initializer_list e =
          scalar type in C that is supported by C's equality comparison. *)
       CStar.Op (K.op_of_poly_comp c)
   | ECast (e, t) ->
-      CStar.Cast (mk_expr env false e, mk_type env t)
+      (* When the source is UInt8/UInt16 and contains arithmetic, the widened
+         result may have extra bits. We must mask before casting to a wider
+         type, otherwise the junk bits are preserved. *)
+      begin match e.typ with
+      | TInt w when w = K.UInt8 || w = K.UInt16 ->
+          let e', is_atomic, _ = mk_arith env e in
+          let e' = if is_atomic then e' else mask w e' in
+          CStar.Cast (e', mk_type env t)
+      | _ ->
+          CStar.Cast (mk_expr env false e, mk_type env t)
+      end
   | EAbort (t, s) ->
       let t = match t with Some t -> t | None -> e.typ in
       CStar.EAbort (mk_type env t, Option.value ~default:"" s)
@@ -702,7 +719,16 @@ and mk_stmts env e ret_type =
           | [] -> None
           | _ -> failwith "impossible"
         in
-        env, CStar.Switch (mk_expr env false false e0,
+        (* In C, the scrutinee is widened to at least int. Since U8/U16
+           have defined overflow, we must mask the extra bits away if the
+           scrutinee is of that type. See https://github.com/FStarLang/karamel/issues/707. *)
+        let scrutinee = match e0.typ with
+          | TInt w when w = K.UInt8 || w = K.UInt16 ->
+              let e, is_atomic, _ = mk_arith env e0 in
+              if is_atomic then e else mask w e
+          | _ -> mk_expr env false false e0
+        in
+        env, CStar.Switch (scrutinee,
           List.map (fun (lid, e) ->
             (match lid with
             | SConstant k -> `Int k
--- a/test/SmallIntCast.fst
+++ b/test/SmallIntCast.fst
@@ -1,0 +1,35 @@
+/// Regression test: ECast on UInt8/UInt16 wrapping arithmetic
+/// doesn't mask before widening, leaking overflow bits.
+module SmallIntCast
+
+open FStar.UInt8
+module U32 = FStar.UInt32
+module Cast = FStar.Int.Cast
+
+/// Cast wrapping addition to UInt32.
+/// F*: (200 + 56) mod 256 = 0, cast to UInt32 => 0ul.
+let cast_add_to_u32 (a b: UInt8.t) : U32.t =
+  Cast.uint8_to_uint32 (a +%^ b)
+
+/// Cast wrapping subtraction to UInt32.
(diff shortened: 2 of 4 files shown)


lib/AstToCStar.ml
```diff
@@ -363,6 +363,13 @@ and mk_arith env e =
          that every subexpression operates over unsigned int until the final
          cast, or until a mask is needed to preserve semantics. *)
       mk_expr env false false e, true, false (* C++: a constant that is wider than the intended type, but in an initializer list, is ok *)
+  | ETernary (c, t, f) ->
```
protz, line 366: are you expecting things of the form `return e1 + e2 ? e3 : e4`?

lib/AstToCStar.ml
```diff
@@ -363,6 +363,13 @@ and mk_arith env e =
...
+      (* Recurse into branches so that non-atomic arithmetic is not hidden
+         behind a ternary marked as "atomic". *)
+      let c = mk_expr env false false c in
+      let t, a1, w1 = mk_arith env t in
+      let f, a2, w2 = mk_arith env f in
+      CStar.Ternary (c, t, f), a1 && a2, w1 || w2
```
protz, line 372: if one of the branches was widened but not the other, shouldn't we widen both? feels like we should assert that `w1 = w2` or do something

lib/AstToCStar.ml
```diff
@@ -454,7 +461,17 @@ and mk_expr env in_stmt under_initializer_list e =
          scalar type in C that is supported by C's equality comparison. *)
       CStar.Op (K.op_of_poly_comp c)
   | ECast (e, t) ->
-      CStar.Cast (mk_expr env false e, mk_type env t)
+      (* When the source is UInt8/UInt16 and contains arithmetic, the widened
```
protz, line 464: yes

protz (approved):
Makes sense, thanks.

Thinking more about my comment regarding the ternary operator, e.g. `(false ? f () : 1) >> 1`. I think it should be have like a binary operator and I think `w1 || w2` is correct.

lib/AstToCStar.ml
protz: this is a conservative choice so I think it is correct

lib/AstToCStar.ml
mtzguido: You mean with an operation in the condition? I don't think so, they must have bool type.

mtzguido:
I have to say I'm a bit confused about the widening. It's only relevant for C++ initializers if I understand correctly? And it seems like an addition of u8 constants is not considered widened, but the addition would be promoted to int?

This generates a bad diff in Eurydice, I'll look into it.
```
tests>  size_t where_clauses_simple_add_81_6c(Eurydice_arr_96 x, size_t y)
tests>  {
tests> -  return (size_t)x.data[0U] + y + (size_t)1U;
tests> +  return (size_t)(uint32_t)x.data[0U] + y + (size_t)1U;
tests>  }
```

I wonder if a different design for all this would be more desirable, actually: keep all casts explicit during AST->C*, and then do a pass to remove the superfluous ones. With enough rules, I think we could remove all the casts in the expression above.

protz:
Sorry I forgot to catch up on this one.

> I have to say I'm a bit confused about the widening. It's only relevant for C++ initializers if I understand correctly?

Yes it happens when Eurydice compiles in C++ mode and relies on a C++ gadget for the construction of tagged unions. I'm re-reading https://en.cppreference.com/cpp/language/list_initialization and I can't seem to either remember or reproduce the issue, but I know that at the time it was necessary. Maybe this is it:

```c++
#include <inttypes.h>

class S {
  uint32_t x;
  public:
  S(uint32_t x) :x(x) {}
};

int main() {
  uint64_t x = 0;
  auto y = S { x }; // ERROR: narrowing conversion not allowed
}
```

I think the logic is that if this is directly under a context that sets the type (like in an initializer list), there's no need to cast to the destination type, UNLESS we are in that particular C++ case.

> And it seems like an addition of u8 constants is not considered widened, but the addition would be promoted to int?

I think the intent is that yes, if this is an addition, it works differently? The example above succeeds with `1ULL` (or `1ULL+1`) in the initializer list.

> This generates a bad diff in Eurydice, I'll look into it.

Thanks

> I wonder if a different design for all this would be more desirable, actually: keep all casts explicit during AST->C*, and then do a pass to remove the superfluous ones. With enough rules, I think we could remove all the casts in the expression above.

Yes, I'm willing to be convinced? It might be easier to prove the correctness via rewriting rules (rather than with the on-the-fly transformation).

mtzguido:
Thanks for the comments, that helps. I'll probably add some tests for that.

> > I wonder if a different design for all this would be more desirable, actually: keep all casts explicit during AST->C*, and then do a pass to remove the superfluous ones. With enough rules, I think we could remove all the casts in the expression above.
> 
> Yes, I'm willing to be convinced? It might be easier to prove the correctness via rewriting rules (rather than with the on-the-fly transformation).

I did start doing this, with encouraging results, but did not finish yet. I could probably get it done in the next week or so. Would you like to merge this PR meanwhile or should we wait for the alternative?

protz:
Let's just merge this on the basis that we don't want to leave something incorrect in the tree if we have a fix for it.

mtzguido:
https://github.com/AeneasVerif/eurydice/pull/407

PR #742, opened 2026-05-03
Generating C compound assignments (x += 2, etc)

mtzguido:
Directly at the level of C11, with a conservative check that the expression is syntactically equal on the LHS and RHS. This way we will not silently drop any inserted casts. The type promotion semantics for a compound assignment and its plain version also match.

### DIFF
--- a/lib/C11.ml
+++ b/lib/C11.ml
@@ -52,7 +52,7 @@ and expr =
   | Member of expr * expr
   | MemberP of expr * expr
   | Assign of expr * expr
-    (** not covering *=, +=, etc. *)
+  | CompoundAssign of expr * K.op * expr (* x op= y, op is binary. *)
   | Call of expr * expr list
   | Name of ident
   | Cast of type_name * expr
--- a/lib/CStarToC11.ml
+++ b/lib/CStarToC11.ml
@@ -46,10 +46,13 @@ let is_compoundable : K.op -> bool = function
      evaluation *)
 let mk_assign (lhs : C11.expr) (rhs : C11.expr) : C11.expr =
   match rhs with
+  | Op2 (Add, e2, Cast (_, Constant (_, "1")))
   | Op2 (Add, e2, Constant (_, "1")) when e2 = lhs ->
     Op1 (PostIncr, lhs)
+  | Op2 (Add, Cast (_, Constant (_, "1")), e2)
   | Op2 (Add, Constant (_, "1"), e2) when e2 = lhs ->
     Op1 (PostIncr, lhs)
+  | Op2 (Sub, e2, Cast (_, Constant (_, "1")))
   | Op2 (Sub, e2, Constant (_, "1")) when e2 = lhs ->
     Op1 (PostDecr, lhs)
   | Op2 (op, e2, e3) when is_compoundable op && e2 = lhs ->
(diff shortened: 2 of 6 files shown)


protz (commented):
I'm quite keen on keeping C11 as a syntactic tree, and having all meaningful operations performed on CStar or earlier.

Is there a deep reason why you did this at the level of C11?

I think I should add a comment at the top of C11 for LLMs to implement their transformations on CStar or earlier :-)

mtzguido:
> I'm quite keen on keeping C11 as a syntactic tree, and having all meaningful operations performed on CStar or earlier.
> 
> Is there a deep reason why you did this at the level of C11?

I kind of see this a just a syntactic convenience, where instead of emitting `E = E + X` we emit `E += X`, hence the smart constructor mk_assign. It then handles both Assign and BufWrite automatically. We could augment CStar with two compound assignments nodes (for Assign and BufWrite) and do a CStar->CStar pass to add them, but that seems like more code for no gain IMO. (This is also where the detection of ++ and -- was.)

Maybe my comment "directly at the level of C11" was misleading. This happens in the C*->C11 translation, and the C11 tree is augmented with compound assignment nodes which is new syntax.

That said I could move this to AstToCStar. Or maybe keep it here but match on the C* syntax instead of C11?

> I think I should add a comment at the top of C11 for LLMs to implement their transformations on CStar or earlier :-)

This one was all me :^)

protz (approved):
Ok my bad I only read the description not the code.

This is the right way :-).

I am wondering what is the argument for "the lhs is always pure", and I guess from F*, all effectful computations are lifted to let-bindings, and are only inlined away if there's one occurrence; for Eurydice, same deal, I think.

mtzguido:
> I am wondering what is the argument for "the lhs is always pure", and I guess from F*, all effectful computations are lifted to let-bindings, and are only inlined away if there's one occurrence; for Eurydice, same deal, I think.

That's my understanding, but actually I'm not convinced this is still true after all the inlining. I couldn't break it after trying a bit, but I'm not too confident it couldn't be done. I added a purity check.

mtzguido:
> > I am wondering what is the argument for "the lhs is always pure", and I guess from F*, all effectful computations are lifted to let-bindings, and are only inlined away if there's one occurrence; for Eurydice, same deal, I think.
> 
> That's my understanding, but actually I'm not convinced this is still true after all the inlining

Just for posterity, in particular I was worried about something like this:
```fstar
fn ret_p ()
  returns p : ref U32.t
  ensures live p
{
  admit();
}

fn test_lhs_purity ()
{
  open U32;
  let p1 = ret_p();
  let p2 = ret_p();
  p1 := !p2 +%^ 1ul;
  drop_ (live p1);
  drop_ (live p2);
  ();
}
```
it currently generates:
```c
void CompoundAssignment_test_lhs_purity(void)
{
  uint32_t *p1 = ret_p();
  *p1 = *ret_p() + 1U;
}
```
which makes sense, since `=` does not enforce an order between lhs/rhs, but F* does evaluate p1 before. But if we ever generated
```c
  *ret_p() = *ret_p() + 1U;
```
the detection would kick in and change the number of calls. I guess we would only ever do that for pure functions... but it seems wrong to have this syntactic pass affect this.

PR #743, opened 2026-05-16
Introduce ESizeof to InputAST

mtzguido:
This allows F* extraction to generate sizeof expressions instead of having to compute sizes internally (impossible for structs due to padding etc).

I got help from Claude for the translation into CFlat, I think it looks right but worth a look.

This does not add tests, unsure what's a nice way to do that since writing a test in F* implies formalizing sizeof to some extent, which we could do but not sure that's a prerequisite. Kuiper is using this already and working well for the C backend at least.


protz (approved):
Looks good to me, thanks!

PR #744, opened 2026-05-16
Inlining: fix bug about loop guards

mtzguido:
This first fixes a [critical bug](https://github.com/FStarLang/karamel/commit/43eb6206bb1f91f0c7594a6d1e2a6271d50b3ece) and then improves inlining to consider the alternative behavior of if/ternary/match/switch, and also sequences. This second patch is worth a skeptic look to make sure I'm not misunderstanding the use analysis.

The first fix decreases output quality slightly when the expression to inline is in fact pure, I will try to improve that later.

### DIFF
--- a/lib/Simplify.ml
+++ b/lib/Simplify.ml
@@ -89,10 +89,28 @@ class ['self] safe_use lvalue = object (self: 'self)
      *every* path in the control-flow is safe. *)
   method! visit_EIfThenElse env e _ _ = self#sequential env e None
   method! visit_ESwitch env e _ = self#sequential env e None
-  method! visit_EWhile env e _ = self#sequential env e None
-  method! visit_EFor env _ e _ _ _ = self#sequential env e None
   method! visit_EMatch env _ e _ = self#sequential env e None
   method! visit_ESequence env es = self#sequential env (List.hd es) None
+
+  (* Loops are more complicated. We cannot just use self#sequential since we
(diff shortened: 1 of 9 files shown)


mtzguido:
My "improvement" was causing the bad diff, not the actual fix, so I removed it for now.

lib/Simplify.ml
```diff
@@ -89,10 +89,28 @@ class ['self] safe_use lvalue = object (self: 'self)
...
+      }
+
+    inlining x ~> a[i] into the loop body is in general not equivalent: b could
+    alias to a, and writes to a[foo] could change the value of a[i] across
+    iterations.  This also applies to the condition: we cannot inline into the
+    condition of a while loop (See InlineLoopCond.fst testcase). The only thing
+    we do allow is inlining into the initializer of a for loop, since it is only
+    executed once. *)
+  method! visit_EWhile _ _ _ = Unsafe
+  method! visit_EFor env _ e _ _ _ = self#sequential env e None
```
protz: why not Unsafe here too then?

protz (changes requested):

lib/Simplify.ml
mtzguido: `e` is just the initializer, not the guard, so it's evaluated only once

protz: ah yes of course -- might be worth adding a comment then

protz (approved):

